Issue #9, opened 2020-02-19
feat: bootnodes healthcheck CI (& generator/aggregator?)


meowsbits:
Migrated from: https://github.com/etclabscore/multi-geth-fork/issues/141
Original author: @meowsbits

---

Bootnodes are nodes that "cold" clients (protocol providers) use to establish a connection with the peer-to-peer network. Inadequate bootnodes for a network can cause either slow or impossible network connections. 

Bootnode lists are currently hardcoded for each supported network.  But liveness (useability) checks are never systematically undertaken, which can cause p2p issues if the lists get invalidated. The current system for bootnode addition is essentially that a node or nodes are proposed, a developer reviews the proposed enodes by pinging them manually (over an arbitrary and variable time course), and then approves the addition. Same, but backwards, for removal.

This feature prop wants to establish a system for
- [ ] bootnode list health checks to be undertaken systematically via a CI environment, where given parameters, inadequate bootnode lists can be flagged for review, purging of dead nodes, and addition of new ones
- [ ] ideally, an environment which could not only provide a health check, but also provide information that could be used in _the generation_ of healthy lists

This would remove tedious developer burden, reduce connection failures, and, in essence, move toward systematizing variables for a critical first process for all devp2p networks.

meowsbits:
__Challenges__: 

- Affirming the availability of nodes in a decentralized p2p network
  + Reachability (nodes might turn off then on, or might be unreachable due to business)
  + Nearness (nodes must be close enough for requests to survive TTL limits)
  + Friendliness (devp2p protocol does not demand or necessarily incentivize that nodes should actually respond to you)

meowsbits:
Closing via https://github.com/etclabscore/discv4-dns-lists and can use the Go test ref'd in #26 separately if needed.

PR #26, opened 2020-02-19
Feat/bootnode healthcheck test

meowsbits:
Rel #9 (well, the healthcheck part. That issue should break into healthcheck vs. generator).
Rel https://github.com/etclabscore/multi-geth-fork/pull/149

This change set implements a basic ping healthcheck program for default bootnode lists.

---

There are challenges when addressing bootnode validations, all of which have to with what "availability" means and looks like on a decentralized p2p network. This feature - as implemented - doesn't really achieve much in the way of actually addressing those questions; it's more like a minimum-viable proof-of-concept and basic heuristic. 

I'm not sure that it should (yet? ever?) be merged, but can at least serve as some sort of maybe-useful reference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,8 @@ matrix:
         tags: true
       skip_cleanup: true
       tag_name: "$TRAVIS_TAG"
-  - os: linux
+  - if: branch =~ /^((?!test-bootnode-availability).)*$/
+    os: linux
     dist: xenial
     go: 1.13.x
     env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,15 @@ go_import_path: github.com/ethereum/go-ethereum
 sudo: false
 matrix:
   include:
-  - if: branch =~ ^ci-test-bootnode-availability
+  - if: commit_message =~ ci:test-bootnode-availability
     os: linux
     dist: xenial
     go: 1.13.x
+    env:
+    - bootnode-availability
     script:
     - make test-bootnode-availability
-  - if: branch =~ /^((?!ci-test-bootnode-availability).)*$/
-    os: linux
+  - os: linux
     dist: xenial
     go: 1.13.x
     script:
@@ -31,8 +32,7 @@ matrix:
         tags: true
       skip_cleanup: true
       tag_name: "$TRAVIS_TAG"
-  - if: branch =~ /^((?!ci-test-bootnode-availability).)*$/
-    os: linux
+  - os: linux
     dist: xenial
     sudo: required
     go: 1.13.x
@@ -63,8 +63,7 @@ matrix:
         tags: true
       skip_cleanup: true
       tag_name: "$TRAVIS_TAG"
-  - if: branch =~ /^((?!ci-test-bootnode-availability).)*$/
-    os: osx
+  - os: osx
     go: 1.13.x
     script:
     - echo "Increase the maximum number of open file descriptors on macOS"
@@ -91,8 +90,7 @@ matrix:
         tags: true
       skip_cleanup: true
       tag_name: "$TRAVIS_TAG"
-  - if: branch =~ /^((?!ci-test-bootnode-availability).)*$/
-    os: linux
+  - os: linux
     dist: xenial
     go: 1.13.x
     env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,20 +3,16 @@ go_import_path: github.com/ethereum/go-ethereum
 sudo: false
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


eyfl (commented):
If any bootnode is temporary offline or availabe in particular region for example China,this check will break the build. So for this point, I'd like to do something anti-engineering, run the shell script mannually.
Add a bootnode, even if it's not reachable for some nodes, it's not a big deal, right?

meowsbits:
Agreed; I don't think we should merge this test at this point. It's too fragile.

Perhaps still valuable as an external tool to use manually, but should not be a CI dependency.

meowsbits:
I've pursued this a little further.

While the test itself is limited (see #9), it still provides _a_ heuristic for developers and maintainers, albeit with imperfect inference. This heuristic just gets a little better when it's at least publicly available (cite-able, requires less/different trust, reproduceable).

These latest changes create a CI/test scheme where the command `make test-bootnode-availability` is run independently on the CI when the branch begins with `ci-test-bootnode-availaibilty`, and never run otherwise. See #31 for an example of this branch/test pattern.

This would let us at least present this heuristic publicly, and skip it's fragility when it's not useful.

The actual `.travis.yml` modification use `if: [blah blah]` and are not super pretty because the logic requires to assert NOT for all test matrix jobs which don't care about this test.  I dislike ugly CI configs as much as anybody, but also think that (a) they're really incidental to the actual code, of which bootnodes are a critical part, and (b) I/we don't ever actually have to gaze upon their ugliness that much.

meowsbits:
Update: with https://github.com/etclabscore/core-geth/pull/26/commits/b865ffe9054dbb628faf6604ec457029bef3eeb3, the CI will now trigger the `bootnode-availability` test only when `ci:test-bootnode-availability` is included anywhere in the "unwrapped" commit message (title + body).

How it works:

The commit message:
```
    .travis.yml: revise CI to run on commit msg regex
    
    Instead of branch prefix regex.
    This gives a little more leeway and configurability,
    and removes some the ugly extaneous if: conditions in the config.
    
    Use like:
    
    ci:test-bootnode-availability
```

contains `ci:test-bootnode-availability`, which fired https://travis-ci.org/etclabscore/core-geth/jobs/653062877?utm_medium=notification&utm_source=github_status

This is intended to remove the rigidity that came with the branch-based pattern described above.

meowsbits:
@tzdybal Curious your thoughts here.

integration/bootnodes_test.go
```diff
@@ -0,0 +1,137 @@
...
+	minPassN := float64(total) * minPassRate // Minimum number of nodes that must be reachable for the test not to fail.
+
+	// Case where pass rate is epsilon (but non-zero) and rounding causes n nodes == 0; infer that at least just 1 node must pass.
+	if minPassN == 0 && minPassRate > 0 {
+		minPassN = 1
+	}
+
+	disc := startV4(t)
+
+nodesloop:
```
tzdybal: I hate labels. I hate "continue nastyLabel" statements. Simple refactor - extract function that executes loop from lines 68-75, but instead of log+continue, just returns RTT and error. And log/count failures depending on return value.
```
unc checkENode(disc *discover.UDPv4, en *enode.Node, maxTrials int) (time.Duration, error) {
       var err error
       for i := 1; i <= maxTrials; i++ {
               start := time.Now()
               err = disc.Ping(en)
               if err == nil {
                       return time.Since(start), nil
               }
       }
       return 0, err
}
```

.travis.yml
```diff
@@ -3,6 +3,14 @@ go_import_path: github.com/ethereum/go-ethereum
 sudo: false
 matrix:
   include:
+  - if: commit_message =~ ci:test-bootnode-availability
```
tzdybal, line 6: TBH I don't see a reason to run this on CI.

eyfl (approved):
LGTM

.travis.yml
meowsbits: My reasoning for it:  If it's a "test" (or heuristic-generating tool) is something we're ever going to use as developers manually, we should do what we can to make the it (and comparable results) publicly available.

> This heuristic just gets a little better when it's at least publicly available (cite-able, requires less/different trust, reproduceable).

integration/bootnodes_test.go
meowsbits: https://github.com/etclabscore/core-geth/pull/26/commits/a542156c9d7c37cda7d240c901fbcadff11ffae3

meowsbits:
Closing this as superseded by #41

Issue #35, opened 2020-02-21
"re-correct" EIP2200 (SLOAD gas does not shift to 800 via this EIP) 


meowsbits:
[EIP2200](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2200.md) was "fixed" with https://github.com/etclabscore/multi-geth-fork/pull/144 (see documented refs there).

But, as it would appear now, EIP2200 _does not actually specify `SLOAD` gas to `800`_; rather, it depends on EIP1884, which does makes that specification. Instead, EIP2200 (silently, not explicitly in its specification) _requires_ EIP1884 as a precondition for algorithmic soundness.

This means that _by the hard spec_ the "original" etclabscore/multi-geth(-fork) implementation was actually correct.

This will also invalidate EIP1086, which was originally documented as a PR here: https://github.com/etclabscore/multi-geth-fork/pull/151, and merged to this codebase here https://github.com/etclabscore/core-geth/commit/57eef090799fa8bd53950779ae9238425d479d73. With this "retrospective allowance" becoming nullified, those changes should at least be un-enabled if not un-implemented totally.

To capture the (technically undocumented, basically just on twitter) intention of EIP2200, the code should implement logic which requires EIP1884 as a condition of EIP2200 enablement. (This should be it's own issue and change set.)

> Applying 2200 without 1884 was never intended and is in fact buggy (SLOAD/SSTORE gas costs are quite messed up in that situation)
> https://github.com/ethereum/EIPs/pull/2514#issuecomment-589190980
![2020-02-21-075845_921x245_screenshot](https://user-images.githubusercontent.com/45600330/75036496-3d9ca300-546f-11ea-88b9-bb6442ed8176.png)

meowsbits:
Fixed with #36

PR #36, opened 2020-02-21
Fix/eip2200 sload gas again (and ECIP1086 nullified)

meowsbits:
Rel #35, and would leave the only thing left to do for that issue to _possibly_ implement EIP1884 as a requirement of EIP2200 (which is still debatable).

- Reverts implementation of EIP2200 setting `SLOAD` gas=800.
- Reverts implementation of EIP1086 which "allowed" EIP2200 to continue with `SLOAD` gas=200.

With this change, Kotti and Mordor testnets will be able to continue as-is (and no other networks supported by default will be broken either).


tzdybal (approved):
LGTM

PR #40, opened 2020-02-25
README.md:update readme file 

eyfl:
Update the description in readme file to align with the new brand

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,26 +22,26 @@ Upstream development from [ethereum/go-ethereum](https://github.com/ethereum/go-
 
 Networks supported by the respective go-ethereum packaged `geth` client.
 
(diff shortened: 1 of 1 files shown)


meowsbits (approved):
LGTM. 

Maybe we should consider simply dropping the `etclabscore/go-ethereum` column since it is indeed obsolete, and thus a comparison against it obsolete as well ?

eyfl:
> LGTM.
> 
> Maybe we should consider simply dropping the `etclabscore/go-ethereum` column since it is indeed obsolete, and thus a comparison against it obsolete as well ?

make sense, `etclabscore/go-ethereum`  got removed.

meowsbits (commented):
Some of the badges on top haven't been updated:
- `API Reference` ref
- `Gitter` ref (we don't have a Gitter set up? maybe should do that...?)
- `Code Shelter` - doesn't apply here

eyfl:
Per @meowsbits 's suggestion, create new Gitter community for core-geth, also update  API reference

Issue #39, opened 2020-02-25
set up ETC DNS discovery


meowsbits:
[ethereum/go-ethereum from v1.9.11](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.11) implements DNS-based peer discovery, currently only supporting Foundation networks.

This ticket is for implementing the same for Classic.

Rel #37

q9f:
Oh, yes please!

meowsbits:
https://github.com/etclabscore/discv4-dns-lists

am running a test cycle now and will push there when aok.

meowsbits:
am running a test cycle now and will push there when aok.

meowsbits:
Finally merged with #47

PR #41, opened 2020-02-25
discovery: add ETC support for discv4 DNS records

meowsbits:
rel #39 

Adds config defaults for nodeset command which allows filtering master ENR list for ETC networks.

Note that this change set uses the branch merging ethereum/go-ethereum@v1.9.11 as a base; that release is where this functionality is introduced.


meowsbits:
Resolves #39 now.

Records are documented and will be updated here https://github.com/etclabscore/discv4-dns-lists

I have a daemon running on a server handling the crawling and list updating.

tzdybal (commented):
I can't accept after merge, but I definitely can say: LGTM.

PR #49, opened 2020-03-02
params: update mordor bootnode list

q9f:
we'll need this for recovering mordor.

- added 6 new bootnodes by q9f
- added 3 new bootnodes by etc coop
- removed bootnodes with private ips (facepalm)
- shuffled the list


meowsbits (commented):
Why shuffled?

meowsbits (approved):

q9f:
I had a hunch after maintaining a dozen nodes at home, the order of the provided nodes might disrupt the overall efficiency of the list. So I randomized the list across all clients.

PR #60, opened 2020-03-06
Feat/docs

meowsbits:
Improving documentation. Ref #61.


meowsbits:
In general, a part of a larger idea to build and coerce docs for core-geth in a better way.  Maybe that can be shared with the core-geth.org website, etc...

I moved a lot of the content from the README into the ./docs/core folder for now, into something that might approximate sections of a webpage or pages of a gitbook.  Most if not all of this content moved is upstream-duplicating, could use a freshen.

Added a focus on How to get started to README, including Docker stuff moar now. Fixed a few links. Updated the comparison table. 

---

docs/core/alltools.md
```diff
@@ -0,0 +1,15 @@
...
+| :----------:  | -------------                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| **`geth`**    | Our main Ethereum CLI client. It is the entry point into the Ethereum network (main-, test- or private net), capable of running as a full node (default), archive node (retaining all historical state) or a light node (retrieving data live). It can be used by other processes as a gateway into the Ethereum network via JSON RPC endpoints exposed on top of HTTP, WebSocket and/or IPC transports. `geth --help` and the [CLI Wiki page](https://github.com/ethereum/go-ethereum/wiki/Command-Line-Options) for command line options.          |
```
tzdybal, line 8: Is the link correct? What about options not supported by EF, like `--classic`?

docs/core/alltools.md
```diff
@@ -0,0 +1,15 @@
...
+| `rlpdump`     | Developer utility tool to convert binary RLP ([Recursive Length Prefix](https://github.com/ethereum/wiki/wiki/RLP)) dumps (data encoding used by the Ethereum protocol both network as well as consensus wise) to user-friendlier hierarchical representation (e.g. `rlpdump --hex CE0183FFFFFFC4C304050583616263`).                                                                                                                                                                                                                                 |
+| `puppeth`     | a CLI wizard that aids in creating a new Ethereum network.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
+
```
tzdybal, line 15: We should at least mention `echaindb` and `echainspec`.

docs/core/run-cli.md
```diff
@@ -0,0 +1,95 @@
+## Running `geth`
+
+Going through all the possible command line flags is out of scope here (please consult our
+[CLI Wiki page](https://github.com/ethereum/go-ethereum/wiki/Command-Line-Options)),
```
tzdybal, line 4: Again, this link won't include `--classic`.

docs/core/run-cli.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Fast node on an Ethereum network
+
+By far the most common scenario is people wanting to simply interact with the Ethereum
+network: create accounts; transfer funds; deploy and interact with contracts. For this
+particular use-case the user doesn't care about years-old historical data, so we can
+fast-sync quickly to the current state of the network. To do so:
+
+```
+$ geth [|--classic|--social|--ethersocial|--mix|--music|--testnet|--rinkeby|--kotti|--goerli|--mordor] console
```
tzdybal, line 16: First `|` unnecessary.

docs/core/run-cli.md
```diff
@@ -0,0 +1,95 @@
...
+the main network, but with play-Ether only.
+
+```shell
+$ geth --testnet console
+```
+
+The `console` subcommand has the exact same meaning as above and they are equally
+useful on the testnet too. Please see above for their explanations if you've skipped here.
+
+Specifying the `--testnet` flag, however, will reconfigure your `geth` instance a bit:
```
tzdybal, line 44: Maybe we should use `--mordor` instead of `--testnet`? But on the other hand, `testnet` is more expressive ;)

tzdybal (commented):
In general, I'm thinking about changing or maybe adding / mentioning ETC counterparts of ETH ecosystem. For example docs can use `--mordor` instead of Rinkeby. And we need to duplicate+modify some parts of EF documentation (like CLI usage).

meowsbits:
@tzdybal Agreed on all points; these are all related to content which would now be in `docs/core/`, which at this point is just a place to "hide" content that I've literally just copy-pasted from the README as-was. Would like to address all your points, but maybe we can make that a next PR?

docs/core/run-cli.md
meowsbits: (Was signifying that no flag is ok too; an optional `any of`.)

tzdybal (approved):
Sure.

eyfl (approved):
LGTM

Issue #63, opened 2020-03-09
TRACKING: Implement available RPC method descriptions for OpenRPC document


meowsbits:
This is a tracking issue for this bigger idea. Descriptions should probably be implemented on a per-module basis or something like that.
eg.
- [ ] `admin_`
- [ ] `eth_` [private]
- [ ] `personal_`
- [ ] ...

meowsbits:
Closing this via the merge of #220, which features method descriptions sourced from method comments. Methods missing documentation should be flagged in separate issues.

PR #66, opened 2020-03-12
WIP: Feat/openrpc describe

meowsbits:
Rel #63 

This is a spike to explore runtime generation of an OpenRPC `rpc_discover` document that completely describes geth's available RPC services. Working initially to pattern this closely against geth, and then abstract to a pluggable package that can be integrated with go-ethereum/rpc, and maybe also expose other OpenRPC library services.

Provided temporarily (for development only) in this PR are a script and logs, as well as a copy of the openrpc.json document itself. These together are intended to make changes to the openrpc.json document transparent and to give context for their implementation. The `jj` the script requires is https://github.com/tidwall/jj. 

Included as submodules and forked for development are cornerstone libraries `github.com/alecthomas/jsonschema` and `github.com/gregdhill/go-openrpc`. 

An example of what adding a method looks like: https://github.com/etclabscore/core-geth/commit/7cbfe11432147eb1d6c412bbe997c7ed2137a76a

### Yields

:page_facing_up: `rpc.discover` document: https://gist.github.com/meowsbits/4da4c08765679dac1899543002d1f545. Note that geth uses a temporary `rpc_describe` method in lieu of the eventual `rpc_discover` name. This is in order to avoid the collision with the exising method, which returns the static **ethereum-json-rpc-specification** document.

:eye: Check it out on the **OpenRPC Playground**: https://playground.open-rpc.org/?url=https://gist.githubusercontent.com/meowsbits/4da4c08765679dac1899543002d1f545/raw/spec.json

### Develop

```
> ./.develop/test.sh
```
> https://github.com/etclabscore/core-geth/blob/feat/openrpc-describe/.develop/test.sh


meowsbits:
https://playground.open-rpc.org/?schemaUrl=https://gist.githubusercontent.com/meowsbits/4da4c08765679dac1899543002d1f545/raw/49b7176be028626e4d0d1c54a9d7e6da6646c9de/ethereum-json-rpc.json

.develop/openrpc-validator-cli.js
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env node
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
```
BelfordZ, line 3: is this the result of running typescript compile?

If so, might I suggest that you include the src ts instead of the compiled result?

If this is meant to be maintained as JS, then lets ditch the export and clean this up a little.

.develop/serve.py
```diff
@@ -0,0 +1,61 @@
+import SimpleHTTPServer
```
BelfordZ, line 1: why add the python dep?

heres a popular one: https://www.npmjs.com/package/serve

.develop/spec2.json
```diff
@@ -0,0 +1,5841 @@
+{
```
BelfordZ, line 1: Recommended that you use @open-rpc/examples here instead - this way you will not have to update your tests to know what version of OpenRPC your code works / doesnt work for.

internal/ethapi/api.go
```diff
@@ -73,27 +74,56 @@ func (s *PublicEthereumAPI) ProtocolVersion() hexutil.Uint {
 	return hexutil.Uint(s.b.ProtocolVersion())
 }
 
+//
+// FIXME: Can't get jsonschema oneOf to work right.
+//
+type EthSyncingResult struct {
```
BelfordZ, line 77: Perhaps you can infer this from the contents of the struct?

for example, when you encounter a struct, it  can be any of the following json schema types:
1. object
2. allOf
3. anyOf

As it turns out, any of these can be in effect, just an object. However, you figure out which to use based on:
- if all the keys in the struct are optional, it MAY be an 'anyOf'
- If all the keys in the struct are required, it MAY be an 'allOf'
- if there is some keys required, others not, it MUST be an object.

While following these heuristics only wont give a perfect solution, perhaps its worth a try?

.develop/openrpc-validator-cli.js
meowsbits: Noted. This is/was only in the PR as a shortcut/convenience for reproducing the document validation heuristic, and I've since made [the script just demand that this tool is installed globally](https://github.com/etclabscore/core-geth/blob/feat/openrpc-describe/.develop/test.sh#L42).

.develop/serve.py
meowsbits: This is here so I can serve the generated doc to a local Open-RPC playground for development visibility. I was getting CORS errors with the normal `SimpleHTTPServer` on Firefox otherwise.

.develop/spec2.json
meowsbits: `spec2.json` is a file generated by the tests.

Well noted on the .@examples point, and will do when I start pulling the logic out and building   tests for it in isolation.

meowsbits:
Closing in favor of #90

PR #71, opened 2020-03-16
Merge/go ethereum v1.9.12

meowsbits:
https://github.com/ethereum/go-ethereum/releases/tag/v1.9.12


accounts/scwallet/hub.go
```diff
@@ -153,7 +153,7 @@ func (hub *Hub) setPairing(wallet *Wallet, pairing *smartcardPairing) error {
 
 // NewHub creates a new hardware wallet manager for smartcards.
 func NewHub(daemonPath string, scheme string, datadir string) (*Hub, error) {
-	context, err := pcsc.EstablishContext(pcsc.ScopeSystem)
+	context, err := pcsc.EstablishContext(daemonPath, pcsc.ScopeSystem)
```
tzdybal, line 156: OMG, we lagged with `go-libpcsclight` version for a long time... This was probably a problem just after migration to go modules. (I remember, that in upstream PR, they notified me about this version problem (and it caused this change in API))

tzdybal (approved):
Lift me up :arrow_up:

accounts/scwallet/hub.go
meowsbits: Hm. I've seen this conflict before, so that would make sense.

PR #74, opened 2020-03-18
consensus/clique: snapshot loop condition fix

meowsbits:
Not sure why this was commented.
Maybe a merge error?

---

Compare upstream at https://github.com/ethereum/go-ethereum/blob/master/consensus/clique/clique.go#L354


q9f:
What does it do?

meowsbits:
makes sure that headers are not applied to a nil snapshot

tzdybal (approved):
I love fixes like this - it take hours to debug, and fix is in one line. LGTM.

Issue #70, opened 2020-03-16
Sync is so slow.


lambda2003:
Geth version: v1.11.1
OS & Version: Ubuntu

I tried to download all blocks as full sync. It was fast but sync speed was so slow in some point at 9670000. Am I doing wrong?? Any help/comment will help me a lot. Thank you for your support as always. :)

======================
current block
======================
currentBlock: 9679390,
highestBlock: 9679544,
knownStates: 489292493,
pulledStates: 489292493,
startingBlock: 9679469

======================
Below is my log file.
======================
   76: cumulative: 4977974 gas: 21000 contract: 0x000000000000000000000000                                                       0000000000000000 status: 1 tx: 0x9e800c550ee5dbdaa2d10277e41e8a1e1eca8fce78ffe49                                                       3ba1d7162ab79b4bc logs: [] bloom: 0000000000000000000000000000000000000000000000                                                       00000000000000000000000000000000000000000000000000000000000000000000000000000000                                                       00000000000000000000000000000000000000000000000000000000000000000000000000000000                                                       00000000000000000000000000000000000000000000000000000000000000000000000000000000                                                       00000000000000000000000000000000000000000000000000000000000000000000000000000000                                                       00000000000000000000000000000000000000000000000000000000000000000000000000000000                                                       000000000000000000000000000000000000000000000000000000000000000000 state:
         77: cumulative: 5014563 gas: 36589 contract: 0x000000000000000000000000                                                       0000000000000000 status: 1 tx: 0xfdcd94cc07c31277ab928b274a1e9fc5f06b3aa906fc112                                                       873f6c3f16687315c logs: [0xc028254e70] bloom: 0000000000000010000000000000100000                                                       00000000000000000000000000000000000000000000000000000000000000000000000000000000                                                       00000000000000000000000000000040000008000000000000000000000010000000000000000000                                                       00000000000000000000020000000000000000000000000080001000000000000000000000000000                                                       00000000000000000000000000000000000000000000008000000000000000000200000000000000                                                       00000000000000000000000000000000000002000000000000000000000004000000000000000000                                                       000000000000000000000000000000000000000000000000000000000000000000000000000000 s                                                       tate:
[...]

PR #80, opened 2020-03-23
core: add ETH blacklisted block 9679398

meowsbits:
Reported here: https://github.com/etclabscore/core-geth/issues/70
Logs here: https://gist.github.com/meowsbits/501b1fa5490e1624d952d606a1fc956b


q9f (approved):

meowsbits:
I think @tzdybal may be doing/wanting something with black vs. whitelists. Deferring to him for continuance of this change set.

meowsbits:
Withdrawing this. This block may be bad for one chain, but not another. Whitelists in generally should be avoided.

PR #90, opened 2020-04-10
rpc.discover by reflection

meowsbits:
image

⚡⚡ Check it out on the **OpenRPC Playground**: https://playground.open-rpc.org/?url=https://gist.githubusercontent.com/meowsbits/4da4c08765679dac1899543002d1f545/raw/spec.json ⚡⚡

The document on the left is generated reflectively (reflexively? reflectingly? reflectsively? ... with some meta programming) and available under the `rpc.discover` method. It outputs an [OpenRPC API Service Description](https://open-rpc.org/)  document that is not only correct and complete, but shows you all (and only) the modules and methods you actually have enabled for a given service. 

The current document geth generates is valid according to https://github.com/etclabscore/openrpc-validator-cli (which I also made... with a whopping pile of approximately 3 loc by standing on the shoulders of giant @BelfordZ s  ... and the rest of the @open-rpc team  :green_salad:).  

---

__Reviewers__

Supersedes #66.  At the time of writing, the contents of these branches are nearly identical, except that the commit counts are 9 (here) vs. 110 there. I more-or-less squashed them into what became this branch. The number of loc here is deceptively high. The vast majority of them were a "manual rebase" on ethereum/go-ethereum/rpc and it's tendrils in github.com/ethereum/go-ethereum/node. I've tried to make the commits descriptive; use them!

... But elsewhere... there are some loc to be seen. Most of __the actual work__ for this PR is in other repositories, which are here included as dependencies. These are:
- https://github.com/etclabscore/go-openrpc-reflect: a library using runtime reflection and static analysis (AST) to pull together a dynamic, correct, and complete map of existing golang RPC service implementations, supporting both Go standard library and ethereum/go-ethereum/rpc patterns. (These packages _also_ use reflection to map Go code (methods on structs) to API services, and are same same but different.)
- https://github.com/etclabscore/go-jsonschema-walk: a library that provides a method `WalkDepthFirst` for jsonschema, providing a depth-first walk of the json schema graph, with a parameter to mutate nodes.
- A couple other critical libraries: [github.com/alecthomas/jsonschema](github.com/alecthomas/jsonschema) and [github.com/gregdhill/go-openrpc](github.com/gregdhill/go-openrpc), which provide Go typings for their namesake abstractions. This branch uses forks of these, and both have some changes that would like to be merged to their upstreams.

Those other repos `go-openrpc-reflect` and `go-jsonschema-traverse` are where the good stuff is, and will need as much (or far more...)  review as this PR. 

Not implemented yet (__"deliberate feature sparsity"__.. . aka things that, yes, would be nice, maybe even very nice... but imo aren't _mission critical_.. . or that I haven't figure out how to do yet :wink:):

- [x] The `servers` field in the openrpc document. It's an empty `[]`. We can get this by wrapping the `handler`s for each IPC, HTTP, and WS server in node/node.go. BUT, etclabscore/go-openrpc-reflect doesn't have a method for this yet. Yet.  But servers are a nice-to-have. If you're hitting the `rpc_discover` method on the API then you probably have a server address. Which is pretty much what `servers` will tell you. So. 
- There's a map of "special" types to desired respective json schema outputs in node/openrpc.go. These are types which use hex marshaling and/or are "really complex" (umm... untyped, poorly typed, have krakens in their json marshaling methods, etc..) and/or that are common and for some reason don't relfect well and/or that I installed at the beginning of this journey which I felt... might mean something. These need to be assigned to json schema objects with the stench and sweat of a human hand. No, I don't like it either. Not one bit. Are there any _other_ methods that need the hopefully-not-contagious touch of a human hand? You tell me.

image

- Tests. Mmm.. not so much. Not here at least. Well, not in _co-duh_... but I can tell you by the 250 commits to the test output [spec.json](https://gist.github.com/meowsbits/4da4c08765679dac1899543002d1f545/) and the legibility of https://playground.open-rpc.org/?url=https://gist.githubusercontent.com/meowsbits/4da4c08765679dac1899543002d1f545/raw/spec.json, that I've had some eyesss onnn itttt.  What would be even more nice is to provide at least a test that makes the `rpc_discover` call and successfully intermarshals its result to github.com/gregdhill/go-openrpc, and validates against the nearer-native js-based https://github.com/etclabscore/openrpc-validator-cli. And maybe a few (random?) tests for specific methods' contents (or a templated match), and maybe even schema count (that way if the API changes, we'll have to know that). 

image

- [x]  I know. There is a big `: tcp:127.0.0.1:8545` in the title of the playground documentation. I put it there to signify that we need to install the servers feature at github.com/etclabscore/go-openrpc-reflect.
- What do you think about how I put the timestamp in the document `info.verson` field? (eg. `1.0.10-2020-04-10T15:32:11-05:00-1586550731`) 

image

- Subscriptions. They're on there. But they're not gonna work for you. Probably not. They might. They won't.  Should we just skip them? Probably. This would be a good problem to solve.  `eth.syncing` too at least. Same method name. Different receivers. Nonunique openrpc method names. How to distinguish? Ethereum uses application specific logic for this in its RPC server package, segregating subscriptions from normal callbacks.

__At github.com/etclabscore/go-openrpc-reflect__, there are some things:
- The links (`externalDocs`) to Github original sources that are a part of the default method `description` won't work for forks... like core-geth. They point to github.com/ethereum/go-ethereum because that's (still) the actual name of the package(s) repo'd here.
- The `servers` configuration support as noted above.

Here's __what's different from #66__.

- left out `.develop/`, which contains test scripts, generated data, and logs. They are still very much relevant and usable, and you should feel welcome to use them in review.
- left out a lot of other scratch and stray .json and .logs.
- a few revised types which improved their reflection outcomes, but were not necessary to the implementation
- there was a method name change in mobile/ethclient.go. I left that out too. No idea how it got there.

:calendar: :call_me_hand: It would be good to set up a time to look over this together. (1) Once for code review, and (2) once for the openrpc document itself (any methods we want to strike from the list? any missing? any types whose documentation leaves something to be desired? do we want to pull some more documentation from another source on certain methods? &c.)


meowsbits:
## A reminder/overview of architecture:

.    https://github.com/etclabscore/core-geth (PR here: https://github.com/etclabscore/core-geth/pull/90)
.   .     https://github.com/etclabscore/go-openrpc-reflect: "Reflect an OpenRPC document from Go code"
.   .   .   https://github.com/etclabscore/go-jsonschema-walk: library for walking a json schema, calling a single mutation function on each node
.   .   .   .   (http://github.com/meowsbits/node-walk): a prototype generic walk program with cycle detection and adequate tests
.   https://github.com/etclabscore/json-schema-to-types: used to generate `meta_schema.go` code used in https://github.com/etclabscore/go-openrpc-reflect, providing types (eg. `type MethodObject struct`, `type ContentDescriptorObject struct` for the OpenRPC document that should be reflected from server code.
.   https://github.com/etclabscore/go-openrpc-petstore: a sample application using https://github.com/etclabscore/go-openrpc-reflect to provide a simple, automatic, and extensible  build+runtime `rpc_discover` endpoint for an existing server application.
.   https://github.com/etclabscore/go-openrpc: __fork__ of https://github.com/gregdhill/go-openrpc: contains OpenRPC document typings (as does https://github.com/etclabscore/go-openrpc-reflect) as well as templating for code generation. I've written some examples as contributions to this repository, namely a cli-generator and a proxy server, both written in Go.

One of those examples I spun off into a standalone app: https://github.com/etclabscore/openrpc-cli-generator.  I also made a node CLI tool for openrpc validation:  https://github.com/etclabscore/openrpc-validator-cli (but keep in the DL cuz so did another openrpc contrib... though I can't find the link atm :(.. .
.  :
.   

:dash:

BelfordZ:
epic

shanejonas:
`personal_signTransaction` has `*` in its name

image

meowsbits:
@shanejonas That's indicating that the variable is a pointer.

BelfordZ:
The result should be a JSON value, so I'm not sure how it could be a pointer. Can you explain what you mean?

meowsbits:
IIUC (afk atm) we’re talking about the ‘name’ value for the content descriptors. The reflector thingamajig sets this value by using the field type name if a field value name (eg ‘myResult’)   is unavailable. 

The variable type isnt dirextly applicable to json, no, but i didnt see much downside to keeping context explicit when possible. (Also a pointer should be correlated with oneOf:null/value for Go, for example, a kind of inter-domain legibilty which might have side effect benefits for certain use cases?)

meowsbits:
In case anyone's interested in experimenting with the program here, here's my development script. It just builds and fires up geth, then makes the `rpc_discover` request and validates the response.  You'll need to change the `gethroot` and `validatorroot` values to be respective of your system. You also need to have `http` (httpie) installed. 

```sh
#!/usr/bin/env bash
set -e

make geth
pkill -f ./build/bin/geth || true

./build/bin/geth --port 30313 --datadir=/tmp/gethddd --nodiscover --maxpeers=0 --rpc --rpcapi=admin,debug,eth,ethash,miner,net,personal,rpc,txpool,web3 >/tmp/geth.log 2>&1 &
disown
gethpid=$!

echo "Gethpid: ${gethpid}"

rm -fr /tmp/gethddd
onexit(){
        kill -9 ${gethpid}
        pkill ./build/bin/geth
        rm -fr /tmp/gethddd
}
trap onexit EXIT

# Wait for geth to start up.
echo "Sleeping 3"
sleep 3

# Post rpc.discover request.

gethroot="$HOME/go/src/github.com/ethereum/go-ethereum"
http --json POST http://localhost:8545 id:=$(date +%s) method='rpc_discover' params:='[]' | jj -p 'result' | tee "${gethroot}/spec.json"

# Validate response.

validatorroot="$HOME/dev/etclabscore/openrpc-validator-cli"
node "$validatorroot/build/index.js" "$gethroot/spec.json" || { cat /tmp/geth.log && echo "FAILED" && exit 1; }
```

meowsbits:
Just a follow up on https://github.com/etclabscore/core-geth/pull/90#issuecomment-620268504 for documentation's sake.

The issue raised was the OpenRPC [Content Descriptor ](https://github.com/open-rpc/spec/blob/master/spec.md#content-descriptor-object) names such as `*SignTransactionResult` are awkward (although this example isn't a good one because it's a result, and the following scenario deals exclusively with params). The core reasoning here is that these `name`s are used as values in the case of methods using `by-name` as their `paramStructure`, making name values with extra-alphanumeric values hard to standardize. These `name` values become keys in the params object, and although extra-alphanumeric characters are allowed in JSON objects, it will be more difficult to convince people to teach machines to use them. See https://github.com/open-rpc/spec/pull/277. https://github.com/etclabscore/core-geth/pull/90/commits/003ac3889bbe05af783aeec3e73ff73ef09f4e35 addresses this, adding application-specific logic to geth to clean up this pattern, stripping these undesirable character values.

There is an effort to establish an OpenRPC specification description standard for Ethereum ~~clients~~ protocol providers which would ostensibly include specification for these names. In this event, the names as generated by reflection will need to be fixed to meet the spec (probably) in many cases. This will be able to be done by renaming Go variables to match the spec, adding variables where there are none, or using a custom value map for these cases.

meowsbits:
Superseded by #137.

PR #103, opened 2020-05-20
replace freezer with a custom implementation

zcstarr:
…mplementation that's toggled by flag


zcstarr:
This is just a dummy PR just so we can quickly look at the code and diff it as we work out what's the best approach to adding a more generic and flexible thing to "servicize or decouple" the freezer

meowsbits:
@zcstarr :construction: Just rebased on master, `--force` pushed.

meowsbits:
Just pushed changes which sketch adding support for AWS S3 as the remote freezer.

You'll need to have your AWS SDK credentials set (https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html). In the script below I have creds as:

```
cat ~/.aws/credentials
[developer-s3]
aws_access_key_id = AKxxx
aws_secret_access_key = Dsxxx
```

Note the CLI flag change, now with __`--datadir.ancient.remote=s3`__.

```sh
env AWS_PROFILE=developer-s3 AWS_REGION=us-east-1 ./build/bin/geth \
  --classic \
  --ipcpath=~/.ethereum/geth.ipc \
  --datadir ~/SAMSUNG_T5/e/sandbox/remote-freezer-develop \
  --verbosity 4 \
  --datadir.ancient.remote=s3 \
  --datadir.ancient.remote.namespace=coregeth-classic-cataclysm |& tee geth.log
```

__NOTE__: S3 demands that bucket names are _globally_ unique. So you'll need to change `coregeth-classic-cataclysm` to be your own special snowflake.

---

~~It currently stores the RLP encoded blobs in S3 buckets per "table", eg. `bodies`, `headers`, etc. Bucket names are structured as `<chainID>-<genesisHash[:8]>-<tableName>`,~~ with the block information being stored by number with an `rlp` extension, eg. `1.rlp`. An ugly `index-marker` is also persisted on S3 that (sketchily) represents the number of frozen items. Index marking and metadata in general will want to be thought through and should for now be understood as a TODO.

Eventually caching will be critical for performance, since `Sync()` is designed to do the heavy lifting (eg flushing to disk) and is called accordingly, ~~but right now `AppendAncient` and `Truncate` are directly writing, and reads are likewise done in sync.~~

meowsbits:
To nuke your buckets between development versions:

```sh
> cat ./geth-s3-buckets-nuke.sh
#!/usr/bin/env bash

env AWS_PROFILE=developer-s3 AWS_REGION=us-east-1 aws s3 rb s3://<my-bucket> --force

```

meowsbits:
A few thoughts.

Features for S3 as-is:

- [x] Fix: use just one bucket. By default, S3 accounts are limited to 100 buckets. 
- [ ] Retroactive upload. "I have already synced. I want to use the remote ancient store without syncing all over again."
- [ ] Read-only remote. "I already have one instance using remote ancients. I want to allow another instance to use the same remote ancient store, and it's ok if it's a slave."
- [ ] Configurable bucket settings. "I don't want to use the default bucket name and/or region."
- [ ] Maybe just store entities in JSON? "I want to use my data elsewhere without a lot of hullabaloo."
- [ ] Maybe (optionally) compress the data? "I want to use ancients, but would prefer spending $0.76/month instead of $1.11"

Extensibility-oriented:

- [ ]  Create an RPC server+client (in place of|congruent to) the S3 implementation. It handles _subscriptions only_ (via websockets and unix sockets) and pushes events for write events/methods (eg. `AppendAncient`, `Sync`) and manages requests for read methods (eg. `Ancient`, `HasAncient`) as well. Now you can write your own standalone remote ancient service.

meowsbits:
So, as-is with the current schema, which adheres strictly to the _freezerTable_ schema (hashes, headers, bodies...) and uses block number as the index, I'm seeing about 2 blocks/second upload, which will take about 2 months to upload for the classic chain's current height. Further, creating 50million relatively tiny objects may have negative implications for the S3 request costs, where I expect fewer, larger objects will be cheaper.

IMO this is beyond usable for most cases, so we need to find a way to make uploads more efficient. 

My understanding so far is that most of our upload time is due to transaction overhead (even though we're `batch`ing uploads by up to 2048 blocks). The _upside_ of block/kind-atomic resources is that we get to download on-demand just what we need, which, in my experience trying out the JS console `eth.getBlock` for remote blocks is "fast" as-is (barely noticeable difference from local FS).

Grouping resources by block (instead of by kind/block) saw about 8 blocks/second upload. I'm going to try reverting to this pattern, but then extending it further to use 32-block objects, with each containing all the tables. If this moves the upload time from 2 blocks/second to `2*4*32 = 256 blocks/second` then I'm going to call that good enough. The downside is that download will take longer, and will download stuff that isn't necessary. 

There is probably a further way of managing this by using an AWS Lambda to partition or group the objects once they're uploaded, but I'm going to call that out-of-scope for now.

As-is:

2020-06-08-211829_771x684_screenshot
2020-06-08-211855_764x994_screenshot

meowsbits:
image

meowsbits:
image

meowsbits:
Just thinking out loud here. I'm coming back to 1 block objects. This keeps GETs streamlined. Will cost about 50USD at the without-any-batching ceiling to set up (upload entire ancient store), mostly in transaction counts. Batching requests in 2048-block chunks (normal during sync) will bring the number down to 10M / 2048 =~ 5k, costing just a few cents in PUT requests. 

The downside still is time, where I'm estimating 2-3 weeks for an upload-enabled sync. Although I'd consider this on the upper end of desirable, I'm currently of the opinion that it is tolerable. Once a user has a canonical bucket uploaded, they can use `s3 sync` to copy entire buckets.

zcstarr:
This is awesome, 

> So, as-is with the current schema, which adheres strictly to the _freezerTable_ schema (hashes, headers, bodies...) and uses block number as the index, I'm seeing about 2 blocks/second upload, which will take about 2 months to upload for the classic chain's current height. Further, creating 50million relatively tiny objects may have negative implications for the S3 request costs, where I expect fewer, larger objects will be cheaper.
> 
> IMO this is beyond usable for most cases, so we need to find a way to make uploads more efficient.
> 
> My understanding so far is that most of our upload time is due to transaction overhead (even though we're `batch`ing uploads by up to 2048 blocks). The _upside_ of block/kind-atomic resources is that we get to download on-demand just what we need, which, in my experience trying out the JS console `eth.getBlock` for remote blocks is "fast" as-is (barely noticeable difference from local FS).
> 
> Grouping resources by block (instead of by kind/block) saw about 8 blocks/second upload. I'm going to try reverting to this pattern, but then extending it further to use 32-block objects, with each containing all the tables. If this moves the upload time from 2 blocks/second to `2*4*32 = 256 blocks/second` then I'm going to call that good enough. The downside is that download will take longer, and will download stuff that isn't necessary.
> 
> There is probably a further way of managing this by using an AWS Lambda to partition or group the objects once they're uploaded, but I'm going to call that out-of-scope for now.
> 
> As-is:
> 
> 2020-06-08-211829_771x684_screenshot
> 2020-06-08-211855_764x994_screenshot

I guess what's behind the slow block upload speed, is it because it seeks on s3 prior to upload? Or is it simply not using a persistent connection to upload?

meowsbits:
> I guess what's behind the slow block upload speed, is it because it seeks on s3 prior to upload? Or is it simply not using a persistent connection to upload?

To be honest I'm not exactly sure. My googling and esoteric forum lurking lead me to "request transaction overhead"-type explanations. Batching (the new SDK API) helps a lot, I'm sure, and although we're typically (during sync) rolling 2048 blocks into a batch, I'm not sure that we're actually making the most of that optimization opportunity. As-is, we are maintaining a single `Session`, as well as `uploader` and `downloader` instances. 

There may also be "transaction overhead" beyond the simple HTTP requests, ie something on the Amazon end with what it costs them (transaction-wise) to create objects -- and we're creating a significant number of rather small objects. They charge per-PUT, so I have to think that although it's in USD "cheap"(-ish), it may speak to the apparent slowness.

In general my intuition is that S3 for millions of tiny things is probably less ideal than S3 for hundreds of thousands of medium-sized things, etc. 

With that said, still, I am not completely satisfied with it as an explanation for why 10M small objects take 2-3 weeks to upload to S3 (although one will find horror stories of long waits to _delete_ large numbers of objects from buckets in the forums). Further investigation might want to create an isolated test case that's totally separate from this blockchain implementation and just try fiddling around with configurations and implementation to get a feel if it's our implementation or if it's just gonna be something like this no matter what.

meowsbits:
last 2 commits fix:

```
anic: runtime error: slice bounds out of range [11:10]

goroutine 384 [running]:
github.com/ethereum/go-ethereum/core/rawdb.(*freezerRemoteS3).TruncateAncients(0xc0001663c0, 0x1, 0x0, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/core/rawdb/freezer_remote_s3.go:337 +0xc0f
github.com/ethereum/go-ethereum/core/rawdb.(*freezerRemote).TruncateAncients(0xc0005b4000, 0x1, 0x0, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/core/rawdb/freezer_remote.go:171 +0x4a
github.com/ethereum/go-ethereum/core.(*BlockChain).truncateAncient(0xc02639e000, 0x0, 0x52d9c50, 0xc021120ee8)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1016 +0x81
github.com/ethereum/go-ethereum/core.(*BlockChain).InsertReceiptChain.func2.1(0xc0211211f0, 0xc02639e000)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1108 +0x60
github.com/ethereum/go-ethereum/core.(*BlockChain).InsertReceiptChain.func2(0xc0263bf850, 0x2, 0x2, 0xc08ea77a70, 0x2, 0x2, 0x0, 0x6f5c930, 0xc09574c380)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1215 +0x1fa7
github.com/ethereum/go-ethereum/core.(*BlockChain).InsertReceiptChain(0xc02639e000, 0xc0263bf840, 0x2, 0x2, 0xc08ea77a40, 0x2, 0x2, 0x9fdae8, 0x0, 0x0, ...)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1305 +0x1d0c
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).commitFastSyncData(0xc0248981c0, 0xc0263bf790, 0x2, 0x2, 0xc0002feb60, 0xc0263bf790, 0x2)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:1755 +0x68c
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).processFastSyncContent(0xc0248981c0, 0xc020a90240, 0x0, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:1678 +0x351
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).syncWithPeer.func7(0xc00014b770, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:529 +0x33
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).spawnSync.func1(0xc0248981c0, 0xc0211ce9c0, 0xc0212801a0)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:543 +0x63
created by github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).spawnSync
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:543 +0xaf
```

core/rawdb/freezer_remote_s3.go
```diff
@@ -0,0 +1,438 @@
...
+
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	n := atomic.LoadUint64(f.frozen)
+
+	foundPending := false
+	f.pendingMu.Lock()
+
+	for k := range f.pending {
```
zcstarr: This logic is a little strange to me it seems like the f.backlogUploads and pending have room to get out of sync?

core/rawdb/freezer_remote_s3.go
```diff
@@ -0,0 +1,438 @@
...
+		// this value indicate outside of range since length is 0 indexed.
+		if num >= items {
+			foundPending = true
+			delete(f.pending, k)
+		}
+	}
+	for i, v := range f.backlogUploads {
+		_, num := fromAwsKey(*v.Object.Key)
+		if num >= items {
+			if i == len(f.backlogUploads) - 1 {
```
zcstarr: something is also a little wonky here. I think I hit the same 
```
panic: runtime error: slice bounds out of range [11:10]

goroutine 377 [running]:
github.com/ethereum/go-ethereum/core/rawdb.(*freezerRemoteS3).TruncateAncients(0xc0001c4240, 0x1, 0x0, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/core/rawdb/freezer_remote_s3.go:342 +0xd3d
github.com/ethereum/go-ethereum/core/rawdb.(*freezerRemote).TruncateAncients(0xc000578000, 0x1, 0x0, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/core/rawdb/freezer_remote.go:171 +0x4a
github.com/ethereum/go-ethereum/core.(*BlockChain).truncateAncient(0xc006944000, 0x0, 0x5306540, 0xc002ca2ec8)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1022 +0x81
github.com/ethereum/go-ethereum/core.(*BlockChain).InsertReceiptChain.func2.1(0xc002ca31d0, 0xc006944000)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1114 +0x60
github.com/ethereum/go-ethereum/core.(*BlockChain).InsertReceiptChain.func2(0xc0000e4940, 0x2, 0x2, 0xc005019560, 0x2, 0x2, 0x0, 0x6b90780, 0xc006780c00)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1221 +0x1faa
github.com/ethereum/go-ethereum/core.(*BlockChain).InsertReceiptChain(0xc006944000, 0xc0000e4930, 0x2, 0x2, 0xc005019530, 0x2, 0x2, 0x9fdf36, 0x0, 0x0, ...)
        /Users/zcstarr/src/etc_labs/core-geth/core/blockchain.go:1311 +0x1d0c
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).commitFastSyncData(0xc004c3c000, 0xc0068d85a0, 0x2, 0x2, 0xc0027b63f0, 0xc0068d85a0, 0x2)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:1770 +0x68c
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).processFastSyncContent(0xc004c3c000, 0xc0002c8b40, 0x0, 0x0)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:1693 +0x351
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).syncWithPeer.func7(0xc000a4d770, 0x493ca2e)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:545 +0x33
github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).spawnSync.func1(0xc004c3c000, 0xc002803620, 0xc0028a7620)
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:559 +0x63
created by github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).spawnSync
        /Users/zcstarr/src/etc_labs/core-geth/eth/downloader/downloader.go:559 +0xaf
```
Issue as aforementioned

zcstarr: Having a bit of an issue decoupling what the intent is, I think there be dragons, because len(mutates) with every append and changes he bounds of the iteration, because it's overwritten.  That said it makes it harder to work out the intent.

core/rawdb/freezer_remote_s3.go
```diff
@@ -334,7 +334,11 @@ func (f *freezerRemoteS3) TruncateAncients(items uint64) error {
 	for i, v := range f.backlogUploads {
 		_, num := fromAwsKey(*v.Object.Key)
 		if num >= items {
-			f.backlogUploads = append(f.backlogUploads[:i], f.backlogUploads[i+1:]...)
+			if i == len(f.backlogUploads) {
```
zcstarr: The issue still persists with this commit. 🤔

meowsbits:
Worth noting that AWS S3 Region chosen will also matter a lot, as well as internet connection (for example, AWS seems to not like some of my VPN endpoints).

meowsbits:
With [latest push](https://github.com/etclabscore/core-geth/pull/103/commits/857a4c8fda9f755bb97dd1af86316397ce2e5a21) I'm getting around 150-200 blocks/second. 

Latest push now uses 32-block groups for object storage. For example, object `blocks/000000032.json` contains blocks 32-63, encoded in an ordered JSON array. 

And I turned my VPN off, so my internet is now on the same continent as my AWS bucket region. 

image
image

And initialization from S3 freezer seems to work (nuke your local db between runs):

```
INFO [06-11|13:36:06.924] Starting Geth on Ethereum Classic...
INFO [06-11|13:36:06.924] Bumping default cache on mainnet         provided=1024 updated=4096
INFO [06-11|13:36:06.927] Maximum peer count                       ETH=50 LES=0 total=50
INFO [06-11|13:36:06.927] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [06-11|13:36:06.950] Starting peer-to-peer node               instance=CoreGeth/v1.11.6-unstable-984826f4-20200611/linux-amd64/go1.14.3
INFO [06-11|13:36:06.950] Allocated trie memory caches             clean=614.00MiB dirty=1024.00MiB
INFO [06-11|13:36:06.950] Allocated cache and file handles         database=/home/ia/SAMSUNG_T5/e/sandbox/remote-freezer-develop2/geth/chaindata cache=2.00GiB handles=524288
INFO [06-11|13:36:07.039] New remote freezer                       freezer=s3 namespace=coregeth-classic-parasail
INFO [06-11|13:36:07.039] New session                              remote=s3 region=0xc000206440
INFO [06-11|13:36:07.039] Creating bucket if not exists            remote=s3 name=coregeth-classic-parasail
INFO [06-11|13:36:07.416] Bucket created                           remote=s3 name=coregeth-classic-parasail result="{\n  Location: \"/coregeth-classic-parasail\"\n}" elapsed=377.046183ms
INFO [06-11|13:36:07.416] Retrieving ancients number               remote=s3
INFO [06-11|13:36:07.470] Finished retrieving ancients num         remote=s3 n=18579
INFO [06-11|13:36:07.470] Initializing cache                       remote=s3 n=18579
INFO [06-11|13:36:07.573] Finished initializing cache              remote=s3
INFO [06-11|13:36:07.829] Persisted trie from memory database      nodes=12356 size=1.78MiB time=23.038484ms gcnodes=0 gcsize=0.00B gctime=0s livenodes=1 livesize=0.00B
INFO [06-11|13:36:07.966] Persisted trie from memory database      nodes=12356 size=1.78MiB time=37.055909ms gcnodes=0 gcsize=0.00B gctime=0s livenodes=1 livesize=0.00B
INFO [06-11|13:36:07.966] Initialised chain configuration          config="NetworkID: 1, ChainID: 61 Engine: ethash EIP1014: 9573000 EIP1052: 9573000 EIP1108: 10500839 EIP1344: 10500839 EIP140: 8772000 EIP145: 9573000 EIP150: 2500000 EIP152: 10500839 EIP155: 3000000 EIP160$
 3000000 EIP161abc: 8772000 EIP161d: 8772000 EIP170: 8772000 EIP1884: 10500839 EIP198: 8772000 EIP2028: 10500839 EIP211: 8772000 EIP212: 8772000 EIP213: 8772000 EIP214: 8772000 EIP2200: 10500839 EIP658: 8772000 EIP7: 1150000 EthashECIP1010Continue: 5000000 EthashECIP1010Pa$
se: 3000000 EthashECIP1017: 5000000 EthashECIP1041: 5900000 EthashEIP100B: 8772000 EthashEIP2: 1150000 EthashHomestead: 1150000 "
INFO [06-11|13:36:07.966] Disk storage enabled for ethash caches   dir=/home/ia/SAMSUNG_T5/e/sandbox/remote-freezer-develop2/geth/ethash count=3
INFO [06-11|13:36:07.966] Disk storage enabled for ethash DAGs     dir=/home/ia/.ethash count=2
INFO [06-11|13:36:07.966] Initialising Ethereum protocol           versions="[65 64 63]" network=1 dbversion=<nil>
WARN [06-11|13:36:07.966] Upgrade blockchain database version      from=<nil> to=7
INFO [06-11|13:36:08.972] Initializing database from freezer       total=18579 number=416 hash="9132b6…f2b7ac" elapsed=1.005s
INFO [06-11|13:36:17.030] Initializing database from freezer       total=18579 number=3552 hash="1b25ec…f5502f" elapsed=9.063s
INFO [06-11|13:36:25.080] Initializing database from freezer       total=18579 number=6752 hash="60ce83…cf3161" elapsed=17.113s
INFO [06-11|13:36:33.104] Initializing database from freezer       total=18579 number=10144 hash="a6ae29…047958" elapsed=25.137s
INFO [06-11|13:36:41.139] Initializing database from freezer       total=18579 number=13536 hash="3500e9…94829e" elapsed=33.172s
INFO [06-11|13:36:49.180] Initializing database from freezer       total=18579 number=16960 hash="9a211b…a00a2e" elapsed=41.213s
INFO [06-11|13:36:52.666] Initialized database from freezer        blocks=18579 elapsed=44.699s
INFO [06-11|13:36:52.732] Loaded most recent local header          number=18578 hash="1baf58…84084c" td=9204680585820061 age=4y11mo5d
INFO [06-11|13:36:52.732] Loaded most recent local full block      number=0     hash="d4e567…cb8fa3" td=17179869184      age=51y2mo4d
INFO [06-11|13:36:52.732] Loaded most recent local fast block      number=18578 hash="1baf58…84084c" td=9204680585820061 age=4y11mo5d
INFO [06-11|13:36:52.732] Upgrading chain index                    type=bloombits percentage=0
INFO [06-11|13:36:52.732] Indexed transactions                     blocks=1     txs=0 tail=18578 elapsed="123.491µs"
INFO [06-11|13:36:52.732] Regenerated local transaction journal    transactions=0 accounts=0
INFO [06-11|13:36:53.095] Allocated fast sync bloom                size=1.60GiB
INFO [06-11|13:36:53.109] Initialized fast sync bloom              items=12356 errorrate=0.000 elapsed=13.712ms
INFO [06-11|13:36:53.193] New local node record                    seq=1 id=b88c999c5d67e33b ip=127.0.0.1 udp=30303 tcp=30303
INFO [06-11|13:36:53.193] Started P2P networking                   self=enode://90c69ac7118154b1993c0c382991b97466997cf1051df8cb657755eb648507d9f31929315c8d549c4cfaabdfa196c364ca01a0cd15b0293e6b1b027f5f6f2c74@127.0.0.1:30303
INFO [06-11|13:36:53.194] IPC endpoint opened                      url=/home/ia/.ethereum/geth.ipc
INFO [06-11|13:36:53.225] New local node record                    seq=2 id=b88c999c5d67e33b ip=24.207.216.46 udp=30303 tcp=30303
INFO [06-11|13:36:53.226] Mapped network port                      proto=tcp extport=30303 intport=30303 interface=NAT-PMP(192.168.0.1)
INFO [06-11|13:36:53.247] Mapped network port                      proto=udp extport=30303 intport=30303 interface=NAT-PMP(192.168.0.1)
INFO [06-11|13:37:01.211] Block synchronisation started
```

meowsbits:
This happens on restart quite often:

```
INFO [06-11|14:04:58.465] Loaded most recent local fast block      number=0     hash="d4e567…cb8fa3" td=17179869184        age=51y2mo4d
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x1c0 pc=0xbffec2]

goroutine 1 [running]:
github.com/ethereum/go-ethereum/core.(*HeaderChain).SetCurrentHeader(0xc007174510, 0x0)
        /home/ia/go/src/github.com/ethereum/go-ethereum/core/headerchain.go:492 +0xa2
github.com/ethereum/go-ethereum/core.(*BlockChain).Rollback(0xc00723a000, 0xc008b80000, 0x16371, 0x16b00)
        /home/ia/go/src/github.com/ethereum/go-ethereum/core/blockchain.go:982 +0x36a
github.com/ethereum/go-ethereum/core.NewBlockChain(0x18efe60, 0xc000eeeda0, 0xc005502800, 0x1900b80, 0x24876e0, 0x18ed540, 0xc0065e8620, 0x0, 0x0, 0x0, ...)
        /home/ia/go/src/github.com/ethereum/go-ethereum/core/blockchain.go:304 +0x11e0
github.com/ethereum/go-ethereum/eth.New(0xc000328c00, 0xc000471200, 0x1456d00, 0xc000397890, 0x423838)
        /home/ia/go/src/github.com/ethereum/go-ethereum/eth/backend.go:208 +0xa2c
github.com/ethereum/go-ethereum/cmd/utils.RegisterEthService.func2(0xc000328c00, 0xc0001eb650, 0xc000397a10, 0xc000387980, 0x2)
        /home/ia/go/src/github.com/ethereum/go-ethereum/cmd/utils/flags.go:1757 +0x3d
github.com/ethereum/go-ethereum/node.(*Node).Start(0xc0001fa000, 0x0, 0x0)
        /home/ia/go/src/github.com/ethereum/go-ethereum/node/node.go:210 +0x3f8
github.com/ethereum/go-ethereum/cmd/utils.StartNode(0xc0001fa000)
        /home/ia/go/src/github.com/ethereum/go-ethereum/cmd/utils/cmd.go:67 +0x2f
main.startNode(0xc000186160, 0xc0001fa000)
        /home/ia/go/src/github.com/ethereum/go-ethereum/cmd/geth/main.go:398 +0x7e
main.geth(0xc000186160, 0x0, 0x0)
        /home/ia/go/src/github.com/ethereum/go-ethereum/cmd/geth/main.go:386 +0x12a
gopkg.in/urfave/cli%2ev1.HandleAction(0x12a55c0, 0x15dc640, 0xc000186160, 0xc0003a0960, 0x0)
        /home/ia/go/pkg/mod/gopkg.in/urfave/cli.v1@v1.20.0/app.go:490 +0xc8
gopkg.in/urfave/cli%2ev1.(*App).Run(0xc0000be1a0, 0xc0000320c0, 0xc, 0xc, 0x0, 0x0)
        /home/ia/go/pkg/mod/gopkg.in/urfave/cli.v1@v1.20.0/app.go:264 +0x5dd
main.main()
        /home/ia/go/src/github.com/ethereum/go-ethereum/cmd/geth/main.go:282 +0x55
```

zcstarr:
Adding an a comment here. latest commit supports ancient remote as a json-rpc service 
```
make all
AWS_REGION=us-west-1 AWS_PROFILE=developers-s3 build/bin/ancientremote  --rpcport 5000 --datadir.ancient.remote s3 --datadir.ancient.remote.namespace fubu --http.addr 127.0.0.1 --http
```
then  
```
 env AWS_REGION=us-east-1 AWS_PROFILE=developers-s3 ./build/bin/geth   --classic   --ipcpath=~/.ethereum/geth.ipc   --datadir ~/.s3-core-z   --verbosity 4   --datadir.ancient.remote=client --datadir.ancient.remote.namespace zaner
```
The commands on both side need a little tlc, the client side should just take an rpcPort to connect to and a host, the 
server side should do all the heavy lifting. I haven't cleaned up the edges yet, so everything is just hard coded to port 5000, and has some duplicate namespace specification. 

@meowsbits let's sync soon and talk about restructuring the internals.

core/rawdb/schema.go
```diff
@@ -78,30 +78,30 @@ var (
 )
 
 const (
-	// freezerHeaderTable indicates the name of the freezer header table.
-	freezerHeaderTable = "headers"
+	// FreezerHeaderTable indicates the name of the freezer header table.
+	FreezerHeaderTable = "headers"
```
meowsbits: This is coming along! 

Just a thought on exporting these bad boys. In terms of maintaining upstream mergability, I'd _generally_ like to avoid this kind of thing, if possible. 

What if we were to just place an equivalent exported layer on top, maybe something like...

- core/rawdb/schema_remote.go (new file, never has merge conflicts)
```go
const (
  RemoteFreezerHeaderTable = freezerHeaderTable
  RemoteFreezerHashTable = freezerHashTable
  // ...
)
```

With the `Remote` prefix hopefully making it clear that that use case is the reason for it being public.

eth/config.go
```diff
@@ -124,12 +124,12 @@ type Config struct {
...
-	DatabaseCache                  int
-	DatabaseFreezer                string
-	DatabaseFreezerRemote          string
-	DatabaseFreezerRemoteNamespace string
+	SkipBcVersionCheck       bool `toml:"-"`
+	DatabaseHandles          int  `toml:"-"`
+	DatabaseCache            int
+	DatabaseFreezer          string
+	DatabaseFreezerRemote    string
+	DatabaseFreezerRemoteIPC bool
```
meowsbits: Why bool and not a filepath string?

core/rawdb/schema.go
meowsbits: And a git thing, please let's try to break massive renames and other strictly-refactorings into their own commits for the sake of diff-grok.

eth/config.go
zcstarr: bool is now used and DatabaseFreezerRemote is either an ipcPath or httpUrl with separate cmd line flags for IPCPath or Remote URL

cmd/ancientremote/server/server.go
```diff
@@ -88,15 +87,6 @@ func MakeServerConfig(c *cli.Context) AncientServerConfig {
 	}
 }
 
-func checkImplementsRemoteFreezerAPI(rpcAPIs []rpc.API) {
```
zcstarr: I think this is not business logic in this sense. Server represents a framework. External people that use the framework, have no idea about the dynamic requirements of framework. The framework assumes you register at least one api that implements AncientStore.

I.e. a developer creates an ancient remote server using the framework doesn't register anything that implements a freezer api. The service won't work. This check can't be done by server in tests, because server has no knowledge of what a person is or isn't registering.

The check acts as an affordance for the developer. In fact if you don't want to do things in this way, the idea is the developer doesn't use any of the convenience packages for implementation.

Gist is it's just opinionated :)

meowsbits:
I was thinking about how to define the API for the ancient remote server (which is the _expected_ API for the built-in ancient remote client). Leaning to an `interface`, maybe even defined in [./interfaces.go](https://github.com/etclabscore/core-geth/blob/master/interfaces.go).

Thinking 0x-prefixed hex marshaling for RLP data, and "raw" `uint64`s and `string`s otherwise.

core/rawdb/freezer_remote_s3.go
meowsbits: Yea. I'd suggest just trashing the `backlog` and replacing with `lru.Cache`.

cmd/ancientremote/server/server.go
meowsbits: > External people that use the framework, have no idea about the dynamic requirements of framework. The framework assumes you register at least one api that implements AncientStore.

Let's not use the word "framework." It's pretty ambiguous. 

The server _must_ implement a static (not dynamic) API. Geth needs a clear way of defining and communicating the specification of that API (which, thankfully, is very simple!). See https://github.com/etclabscore/core-geth/pull/103#issuecomment-664537116.  With a clearly defined API, we can -- and should -- then include an integration test that can demonstrate a functioning server as a mock for the built in client. 

If a server fails to properly implement the API, then imo geth's client should just fail loudly and immediately when a method call returns an error. There will be different contexts of errors (transport, logic), and at least to begin I would treat them all as `Fatal`. Eventually if we get sophisticated, maybe transport errors can get some try-again leeway or something from geth's client.

meowsbits:
Here's the latest script I use to run `geth` with the preliminary `ancient-remote-s3` server. You'll need to find-and-replace some hardcoded variables specific to my environment.

```sh
#!/usr/bin/env bash

set -e

make all

if [[ "$1" == rm ]]; then
  rm -rf ~/SAMSUNG_T5/e/sandbox/remote-freezer-develop4-mordor
fi

echo "Starting s3-ancient-remote..."
#--bucket=coregeth-classic-pentagon \ # latest --classic dev 4 
#--bucket=coregeth-classic-paradox \
(

pushd /home/ia/dev/etclabscore/ancient-store-s3
go install .
popd

env AWS_PROFILE=developer-s3 AWS_REGION=us-east-1 ancient-store-s3 \
  --bucket=coregeth-mordor-pineapple \
  --ipcpath=/tmp/s3-ancient.ipc >s3-ancient-remote.log 2>&1

)&
s3pid=$!
trap "kill -9 $s3pid" EXIT

# Wait for things to set up
startup_wait=5s
echo "Waiting $startup_wait for s3-ancient-remote to start up..."
sleep $startup_wait

echo "Starting geth..."
(
./build/bin/geth \
  --mordor \
  --ipcpath=~/.ethereum/geth.ipc \
  --ancient.rpc=/tmp/s3-ancient.ipc \
  --datadir ~/SAMSUNG_T5/e/sandbox/remote-freezer-develop4-mordor \
  --verbosity 3 \
  --txlookuplimit 1 \
  --vmodule='core/*=5' >geth.log 2>&1
)&
gethpid=$!
wait $gethpid
```

meowsbits:
Just updated https://github.com/etclabscore/core-geth/pull/103#issuecomment-665271573 with latest script, including latest changes to flag design, ie. `--ancient.rpc`.

https://github.com/etclabscore/core-geth/pull/103/commits/76f8b652d86c5124afb097d380f72cde24e69af9 adds a configurable `GETH_ANCIENT_RPC` env var which can be used to run blockchain tests against a remote freezer server, working as an integration test for both geth and the remote ancient store.

Here's my script for that, using the included memory-backed ancient store command `ancient-store-mem` (a lightweight, ephemeral-only, "mock" (not actually mock, but kind of) server):

```sh
#!/usr/bin/env bash

set -x

make all

echo "Starting ancient-store-mem (ephemeral remote ancient store)..."
(
./build/bin/ancient-store-mem ./ancient.ipc > ancient-store.log 2>&1
)&
aspid=$!
trap "kill -9 $aspid"

echo "Running tests..."
[ ! -S ./ancient.ipc ] && echo "Must have running freezer server with ipc at ancient.ipc" && exit 1
env GETH_ANCIENT_RPC="$(pwd)/ancient.ipc" go test -count=1 -v -run='_RemoteFreezer' ./core/... &> test.out
exited=$?

tail -40 test.out

echo "Finding unique method names called (to show completeness in method exercise)"
cat ancient-store.log | grep 'mock server called' | cut -d'=' -f2 | cut -d' ' -f1 | sort | uniq

[ $exited = 0 ] && echo "PASS" || echo "FAIL"

exit $exited
```

node/node.go
```diff
@@ -644,6 +645,39 @@ func (n *Node) OpenDatabase(name string, cache, handles int, namespace string) (
...
+		return nil, err
+	}
+	switch u.Scheme {
+	case "http", "https":
+	case "ws", "wss":
+	case "stdio":
+	case "":
+		// IPC
+		switch {
+		case freezer == "":
```
meowsbits: I don't know about this. I'm not super comfortable providing this 'hidden' default.

zcstarr: in reality you're always checking for ipc filepath and any format  restriction you put on it is hidden and arb. really

zcstarr: I suppose it's tricky here 🤔

zcstarr: err mis read the code yeah I think we should yank it and fail

meowsbits: Done with https://github.com/etclabscore/core-geth/pull/103/commits/5d48b2dcb13526d436f17e16af17df534a14a5bc

eth/config.go
meowsbits: This has been resolved by simply passing the raw freezer URL through to the RPC package, unmutated and unvalidated. If RPC can't handle it, it'll send the error back up and the user will see it and things will come to a grinding halt. Removes the need for the bool field.

core/rawdb/schema.go
meowsbits: This was resolved (completed) with https://github.com/etclabscore/core-geth/pull/103/commits/685132e050e946bd0f4fd48f5fb8a75831cc0622

meowsbits:
Closed via the merge of https://github.com/etclabscore/core-geth/pull/168

PR #111, opened 2020-06-03
Fix/kotti forkid config

meowsbits:
Kotti chain config, which uses Clique consensus,
rendering these configuration values irrelevant.

The existence of these configuration values causes
the client to generate an invalid forkid (eth/64)
value.

Rel https://github.com/ethereumclassic/ECIPs/pull/323

--- 

I've also included additional test cases as well as the test program used to generate the specification linked above.


tzdybal (approved):
LGTM

PR #112, opened 2020-06-05
Merge/foundation 1.9.15

meowsbits:
~~Keeping ahead of the curve, I've merged ethereum/go-ethereum through [their contemporary master](https://github.com/etclabscore/core-geth/commit/4b2ff1457ac28fb2894485194e0e344e84c2bcd7).~~

Merges through da06519347 ([v1.9.15](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.15)). 

This includes some important and notable changes, especially and specifically:
- YoloV1 testnet support, including EIP2537 (BLS precompiles) (`--yolov1`)
- EIP2315 support (3 additional opcodes)

Note that EIP2315 is not configurable nor supported by the `ChainConfigurator` interface, but is implemented and tested in the core/vm package.

In addition to resolving merge conflicts, I also fixed some issues and chain configuration, especially where available chain configurations were not present but should have been.

I found and fixed a bug in all `ChainConfigurator` interface implementors pertaining to `Ethash`, where ethash-specific features were allowed to return non-nil values when the config's engine was Clique.

PR #114, opened 2020-06-09
main: copydb cmd: remove redefined ropsten flag

meowsbits:
Prevented `--help` from working for the command.


meowsbits:
Fixed at master by #112

PR #115, opened 2020-06-09
main: include missing supported flags for 'geth inspect' command

meowsbits:


meowsbits:
Fixed on master via #112

PR #132, opened 2020-07-01
install regression tests against condensed, canonical-simulating blockchains

meowsbits:
The bulk and meat of this commit is in the export.rlp.gz
files, which contain compressed and encoded blockchain data
of relatively small blockchains using 'condensed' versions
of canonical mainnet configurations for Ethereum Foundation
and Ethereum Classic mainnet chains.

Please see tests/regression/simulated/README.md for more
information.

---

Note that foundation is supported here by 2 tests. These are probably, and hopefully, redundant. However, since the transaction-seeding process is non-deterministic (since it happened on a live, mining blockchain under varying conditions), they are not exactly equivalent. The `-2` version contains slightly more data, since it receive more transactions. They use identical configurations. 

The tests are installed in the Makefile, and are run as a dependent of `test-coregeth`, which also includes other core-geth -specific Go tests.


meowsbits:
Test (ie blockchain) generation processes and programs can be found here: https://github.com/meowsbits/core-geth/blob/feat/classic-squashed-chain-regression-gitsquashed/runminer.sh.

PR #145, opened 2020-07-16
Check if mainnet in a function

richardpringle:
Core-Geth is bumping up the cache size because it doesn't recognize Kotti as being a testnet


meowsbits (approved):
LGTM! Thanks @richardpringle , nice catch.

PR #168, opened 2020-08-04
add support for ancient storage as json-rpc service

zcstarr:
add support for ancient storage as json-rpc service

This change adds support for decoupling ancient storage from
disk to a json-rpc supported service. It defines FreezerRemoteClient
as the interface that services must implement in order to be an
ancient storage provider.

This change will allow developers to build custom long term storage
solutions for ancient blocks.

The protocols supported for a storage provider are as follows,
http,https,ws,wss and ipc.

The remote ancient storage can be accessed via the newly added
ancient.rpc flag provided at the commandline. This may not be used
in conjunction with ancient.datadir


zcstarr:
this PR is the squashed version #103

cmd/ancient-store-mem/lib/mem.go
```diff
@@ -0,0 +1,130 @@
...
+	defer f.mu.Unlock()
+	for i, fv := range fields {
+		kind := fieldNames[i]
+		f.store[f.storeKey(kind, number)] = fv
+	}
+	return nil
+}
+
+func (f *MemFreezerRemoteServerAPI) TruncateAncients(n uint64) error {
+	fmt.Println("mock server called", "method=TruncateAncients")
```
zcstarr, line 121: should we be using log here @meowsbits  ?

core/rawdb/database.go
```diff
@@ -102,6 +102,79 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+func NewDatabaseWithFreezerRemote(db ethdb.KeyValueStore, freezerURL string) (ethdb.Database, error) {
+	// Create the idle freezer instance
+	log.Info("New remote freezer", "freezer", freezerURL)
+
+	frdb, err := newFreezerRemoteClient(freezerURL)
+	if err != nil {
+		log.Error("NewDatabaseWithFreezerRemote error", "error", err)
+		return nil, err
+	}
+	// Since the freezer can be stored separately from the user's key-value database,
```
zcstarr, line 127: We should update this comment to reflect what implementers should do or remove it altogether

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/VictoriaMetrics/fastcache v1.5.7
 	github.com/aristanetworks/goarista v0.0.0-20170210015632-ea17b1a17847
 	github.com/aws/aws-sdk-go v1.25.48
-	github.com/btcsuite/btcd v0.0.0-20171128150713-2e60448ffcc6
+	github.com/btcsuite/btcd v0.20.1-beta
```
zcstarr, line 13: is this version diff problematic @meowsbits ?

go.mod
```diff
@@ -68,12 +68,13 @@ require (
 	github.com/wsddn/go-ecdh v0.0.0-20161211032359-48726bab9208
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
 	golang.org/x/net v0.0.0-20200625001655-4c5254603344 // indirect
-	golang.org/x/sync v0.0.0-20190423024810-112230192c58
+	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
```
zcstarr, line 71: ditto

node/service.go
```diff
@@ -48,6 +48,24 @@ func (ctx *ServiceContext) OpenDatabase(name string, cache int, handles int, nam
...
 }
 
+// OpenDatabaseWithFreezer opens an existing database with the given name (or
+// creates one if no previous can be found) from within the node's data directory,
+// also attaching a chain freezer to it that moves ancient chain data from the
+// database to immutable append-only files. If the node is an ephemeral one, a
+// memory database is returned.
+func (ctx *ServiceContext) OpenDatabaseWithFreezerRemote(name string, cache int, handles int, freezer string) (ethdb.Database, error) {
+
+	//fmt.Println("openning the FREEZA via rawdb p1")
```
zcstarr: rm cleanup

node/service.go
```diff
@@ -48,6 +48,24 @@ func (ctx *ServiceContext) OpenDatabase(name string, cache int, handles int, nam
...
+	if ctx.Config.DataDir == "" {
+		return rawdb.NewMemoryDatabase(), nil
+	}
+
+	root := ctx.Config.ResolvePath(name)
+	//fmt.Println("openning the FREEZA via rawdb p1.5")
```
zcstarr: rm cleanup

zcstarr (commented):
@meowsbits PTAL as well I'll patch up the things I found

node/service.go
meowsbits: https://github.com/etclabscore/core-geth/pull/168/commits/4754e617dd82c87b9932725e40e699633b9eb6ce

node/service.go
meowsbits: https://github.com/etclabscore/core-geth/pull/168/commits/4754e617dd82c87b9932725e40e699633b9eb6ce

go.mod
meowsbits: It doesn't appear problematic to me. Also not sure if relevant or necessary, though.

core/rawdb/database.go
meowsbits: The commented code here differentiates core-geth from go-ethereum a little. Where go-ethereum won't attempt recovery in case of invalidation, core-geth will attempt to fix things. 

Although it would introduce a degree of variance, one idea here is to use the validate-only logic that go-ethereum uses. This will prevent mixing up freezers and KVs, while not implementing "fancy" recovery logic that may or may not be handled well by the remotes.

meowsbits: See https://github.com/etclabscore/core-geth/pull/168/commits/6707b46d3ceb5da6bddb4472c11363519b745767

cmd/ancient-store-mem/main.go
```diff
@@ -0,0 +1,20 @@
+/*
+Copyright © 2020 NAME HERE <EMAIL ADDRESS>
```
meowsbits: FIXME

cmd/ancient-store-mem/main.go
```diff
@@ -0,0 +1,20 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+func main() {
```
meowsbits, line 19: Add `doc.go` or `README.md` to document this program. Ie. 

```
	Use:   "ancient-store-mem",
	Short: "Memory-backed remote ancient store application",
	Long: `Uses a memory-backed map to store ancient data.
This application is intended for testing purposed only.
Ancient data is stored ephemerally.
Expects first and only argument to an IPC path, or, the directory
in which a default 'mock-freezer.ipc' path should be created.
Package 'lib' logic may be imported and used in testing contexts as well.
```

cmd/utils/flags.go
```diff
@@ -1564,6 +1570,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 	// Ancient tx indices pruning is not available for les server now
 	// since light client relies on the server for transaction status query.
 	CheckExclusive(ctx, LegacyLightServFlag, LightServeFlag, TxLookupLimitFlag)
+
+	CheckExclusive(ctx, AncientFlag, AncientRPCFlag)
```
meowsbits, line 1565: Flag exclusivity is only checked here, but ancient flags are available on other commands too.

cmd/utils/flags.go
```diff
@@ -1586,6 +1595,13 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 	if ctx.GlobalIsSet(AncientFlag.Name) {
 		cfg.DatabaseFreezer = ctx.GlobalString(AncientFlag.Name)
 	}
+	if ctx.GlobalIsSet(AncientRPCFlag.Name) {
+		remote := ctx.GlobalString(AncientRPCFlag.Name)
+		if _, err := url.Parse(remote); err != nil {
```
meowsbits: Can we just trust the value fall to `rpc.Dial` and get handled (fail hard, database won't open) if there's an error there?

core/rawdb/freezer.go
```diff
@@ -269,17 +269,24 @@ func (f *freezer) freeze(db ethdb.KeyValueStore) {
...
+
 		// Retrieve the freezing threshold.
 		hash := ReadHeadBlockHash(nfdb)
 		if hash == (common.Hash{}) {
 			log.Debug("Current full block hash unavailable") // new chain, empty database
 			backoff = true
 			continue
 		}
+
+		numFrozen, err := f.Ancients()
```
meowsbits: Just a call for :eyes: . 

Using an assigned value like this rather than having direct access to the 'live' `f.frozen` field is a drastic logical step. We need to be sure that the value will always be updated in sync with the subsequent logic depending on it, and that we're never going to be thrashing the RPC in this loop.

core/rawdb/freezer_remote_client.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, err
+	}
+	return &FreezerRemoteClient{
+		client: client,
+	}, nil
+}
+
+// Close terminates the chain freezer, unmapping all the data files.
+func (api *FreezerRemoteClient) Close() error {
+	api.mu.Lock()
```
meowsbits: Is this mutex reasonable?

core/rawdb/freezer_remote_client_test.go
```diff
@@ -0,0 +1,90 @@
...
+			err := frClient.TruncateAncients(*head)
+			if err != nil {
+				t.Fatalf("truncate: %v", err)
+			}
+		}
+	}
+
+	head := uint64(0)
+	for i := 0; i < 1000; i++ {
+		ancientTestProgram(&head, i)
```
meowsbits, line 88: A final-state check here would be good.

rpc/client_test.go
```diff
@@ -34,6 +35,20 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+func TestParseRawURL(t *testing.T) {
```
meowsbits: DELETEME

meowsbits (commented):
See comments. Some tiny obvious things to clean up. Some to review together.

cmd/utils/flags.go
zcstarr: I think we can resolve this. All commands that refer to ancient flag call SetEthConfig, which in term triggers this flag check ( inspect/copy/import) in fact all geth commands pass through the checks in SetEth.

zcstarr: additionally there weren't any other bins using AncientFlag

meowsbits: Got it, SGTM.

cmd/ancient-store-mem/main.go
meowsbits: I will do this.

core/rawdb/freezer_remote_client_test.go
meowsbits: I will do this.

rpc/client_test.go
meowsbits: I will do this.

cmd/ancient-store-mem/main.go
meowsbits: I will do this.

cmd/ancient-store-mem/main.go
meowsbits: Ah, you did this. https://github.com/etclabscore/core-geth/pull/168/commits/7a7f8d793cbe57d6d99126c7749557029e8551c0

core/rawdb/freezer.go
meowsbits: And this modifies the built-in FS ancient database mechanism. So, double double check.

core/rawdb/freezer_remote_client.go
meowsbits: Since the sibling FS `freezer` implementation _doesn't_ use a mutex around its methods, I think we can safely remove this. If freezer implementations want to implement a mutex at the server level, they can. Call atomicity should be managed by the caller and the implementer, not the middleman.

@zcstarr Do you agree?

core/rawdb/freezer_remote_client_test.go
meowsbits: https://github.com/etclabscore/core-geth/pull/168/commits/2707f0f8ed14324704502b5943e2632b76da9a3d

rpc/client_test.go
meowsbits: https://github.com/etclabscore/core-geth/pull/168/commits/2849633505f328276693dcd6ccb894625e90b114

core/rawdb/freezer_remote_client.go
meowsbits: Did with https://github.com/etclabscore/core-geth/pull/168/commits/3652204011544e7d04bf0ef317a7e5b1ae754a89. 

If you disagree, we'll revert.

cmd/ancient-store-mem/main.go
meowsbits: https://github.com/etclabscore/core-geth/pull/168/commits/2acbe19890cfdf7bfe46da32212fbf01208a9b4c

core/rawdb/freezer_remote_client.go
zcstarr: I approve of this

core/rawdb/freezer.go
meowsbits: An alternative here that may actually be preferable would be to duplicate the logic into

```go
// original, unmodified FS implementation, using f.frozen and unmodified function signature
func (f *freezer) freeze(db ethdb.KeyValueStore) {

// freezer client implementation, using f.Ancients() and modified function signature
func freezeRemote(db ethdb.KeyValueStore, f ethdb.AncientStore, quitChan chan struct{}) {
```

`NewDatabaseWithFreezer` and `NewDatabaseWithFreezerRemote` are the only two callers, so the refactoring and code dependencies are simple.

zcstarr: I agree with that then we can reduce the surface area of things that can go wrong and prove out that ancient remote works as expected over a longer time frame.

meowsbits: https://github.com/etclabscore/core-geth/pull/168/commits/f528c61da5963f9eb26442a88f5ce001fc474097 implements [this proposal :point_up:](https://github.com/etclabscore/core-geth/pull/168#discussion_r465894730). Can easily revert if we decide against.

zcstarr: yeah I think this makes the most sense for now, it gives us some space to prove out the remote freezer logic. any thoughts on keeping track of any critical upstream patches to freezer within remote freezer?

meowsbits: I mean, to me the code before https://github.com/etclabscore/core-geth/commit/f528c61da5963f9eb26442a88f5ce001fc474097 looked OK too (with `f.frozen` using `f.Ancients()` instead). The only reason for the refactoring to near-dupe functions is to reduce the risk in case I'm wrong. As @zcstarr pointed out however, duplication introduces an overhead in code-life audits and maintenance.

zcstarr: for the record here, I'd say the code looks correct for f.Ancients() as well.

meowsbits:
I just tested:
- [x] geth initializes from freezer (having called `geth removedb` to remove state db, ancient at ETC block 10530586)

zcstarr:
final merged changes looked good

PR #170, opened 2020-08-06
install immutability threshold flags

meowsbits:
This is a rush job, but should let users configure immutability thresholds as they see fit:

```
geth --immutabilitythreshold.full=3000
```

```
geth --immutabilitythreshold.light=3000
```


zcstarr (commented):
These changes LGTM 🚢  post merge

PR #180, opened 2020-09-12
Feat/fakepow delaysealer

meowsbits:
Installs a configurable fake delay for the PoW `ModeFake` sealer. 

Sealing time determined by a Poisson distribution and governed by `ethash.threads=<mean seconds (lambda)>`.


meowsbits:
```
TestSealFakeDelay: sealer_test.go:49: ethash.threads,lamda= 3 [1 2 2 2 2 2 2 3 3 3 3 3 3 4 4 4 4 4 5 6]
TestSealFakeDelay: sealer_test.go:49: ethash.threads,lamda=10 [5 6 6 7 8 8 8 9 9 10 10 10 10 11 11 11 12 13 14 16]
TestSealFakeDelay: sealer_test.go:49: ethash.threads,lamda=19 [12 13 13 14 14 14 15 16 17 19 20 20 21 22 23 25 25 25 26 26]
TestSealFakeDelay: sealer_test.go:49: ethash.threads,lamda=33 [19 22 24 26 27 27 28 28 28 31 34 36 36 37 38 39 40 40 40 47]
```
https://github.com/etclabscore/core-geth/pull/180/commits/0177c08c9f545c0f2db252c0fda258fc17fda79f

meowsbits:
Closing this for now since we have no immediate intention of merging the feature to master. 

Rel #188, though, so if pursuing that feature, keep this feature in mind too.

consensus/ethash/sealer.go
```diff
@@ -46,18 +47,53 @@ var (
...
+// makeFakeDelay uses the ethash.threads value as a mean time (lambda)
+// for a Poisson distribution, returning a random value from
+// that discrete function. I think a Poisson distribution probably
+// fairly accurately models real world block times.
+// Note that this is a hacky way to use ethash.threads since
+// lower values will yield faster blocks, but it saves having
+// to add or modify any more code than necessary.
+func (ethash *Ethash) makeFakeDelay() float64 {
+	p := distuv.Poisson{
+		Lambda: float64(ethash.Threads()),
```
meowsbits, line 60: Either should or should not install a randomness source here.

ziogaschr: @meowsbits Can you explain more as I don't understand what you want to be done here.

ziogaschr:
This PR is going to be used for #243, so we are re-opening it

consensus/ethash/sealer.go
meowsbits: Sorry, I was not clear at all :facepalm: 

The `distuv.Poisson` struct exports two fields: `Lamba` and `Src`. 

- https://pkg.go.dev/gonum.org/v1/gonum@v0.8.1/stat/distuv#Poisson
- https://pkg.go.dev/golang.org/x/exp/rand#Source

So we can set a pseudo-random source like `rand.NewSource(time.Now().UnixNano())`. Although I think there are far-edge cases where _not_ having a source would be desirable, I think they're easily outweighed by having one.

meowsbits: Done with https://github.com/etclabscore/core-geth/pull/180/commits/e695a470652d0d73bbf75ddacb3e7d6725aea0c4

cmd/utils/flags.go
```diff
@@ -1469,6 +1469,10 @@ func setEthashCacheDir(ctx *cli.Context, cfg *eth.Config) {
 }
 
 func setEthash(ctx *cli.Context, cfg *eth.Config) {
+	if ctx.GlobalBool(FakePoWFlag.Name) {
```
meowsbits: @ziogaschr What do you think about including this in this PR? 

Although it does indeed seem like something worth doing, I'm inclined to think it should be it's own PR. Maybe we can also think about extending through other possible variations, like maybe `--fakepow.poisson`... 

Thoughts?

ziogaschr: Indeed I would prefer this removed from this PR. It might fit well into #243.
I will remove it now

ziogaschr (approved):

cmd/utils/flags.go
ziogaschr: About the `--fakepow.poisson`, I was thinking something like `--ethash.mode` or `--ethash.powmode` enum string flag with values **fake**, **fullFake**, **poissonFake**.

Also, if we add it as a generic cli command and **not under** the `--dev.*` namespace, I would suggest adding something more fancy like `--ethash.[pow]modedangerously`.

---

Not sure how the **fakepow** will be used, outside of the geth [importChain, exportChain, copyDb, dump, inspect] cmds context, but for now it might make sense to only activating under the `--dev` flag and not on real chains.

PR #185, opened 2020-09-15
rpc: add admin_maxPeers method

meowsbits:
This is a feature that we want for use in testing on the MESS testnet. It has been merged to `rc/messnet-1` branch, and will be included in the `v1.11.14-rc-messnet-2` release.

Don't merge me to master right away with some discussion about whether we actually want to do that beyond the case of this testnet.

PR, for the time being, mostly just for documentation and transparency.


meowsbits:
My reasoning for the proposed addition is that it's a useful thing to have, both for test scenarios and node administration. The drawbacks I see are: it adds a security-sensitive command to an API, and it adds code. The retort to these counterarguments are that: security-sensitive commands are exactly what the `admin_` namespace is for (and there are "worse" things that a nasty guy can do if given access to it), and the code added is pretty self-contained and simple.

ziogaschr (approved):
LGTM

PR #186, opened 2020-09-18
ecip 1099

iquidus:
Implements ECIP-1099 (Calibrate Epoch Duration) specified at https://github.com/ethereumclassic/ECIPs/pull/367


iquidus:
One thing im unsure on is how much we want to ethash > etchash. imo anything additional is primarily cosmetic, and adds burden to future maintenance. i.e renaming the consensus engine.

meowsbits:
Yea, wherever (in the code) the name doesn't serve an important and functional distinction, I would use `ethash`.  After all, core-geth still supports a lot of different networks!

consensus/ethash/algorithm.go
```diff
@@ -39,30 +39,47 @@ const (
 	datasetGrowthBytes = 1 << 23 // Dataset growth per epoch
 	cacheInitBytes     = 1 << 24 // Bytes in cache at genesis
 	cacheGrowthBytes   = 1 << 17 // Cache growth per epoch
-	epochLength        = 30000   // Blocks per epoch
+	oldEpochLength     = 30000   // Blocks per epoch pre ECIP-1099 activation
+	newEpochLength     = 60000   // Blocks per epoch post ECIP-1099 activation
```
meowsbits: `old` and `newEpochLength` at least need to be renamed to explain their association to ECIP1099, since this change is only old/new for Ethereum Classic and her testnets.  Maybe `epochLengthAppendixJ` (or `epochLengthDefault`?) and `epochLengthECIP1099`? 

Got the `AppendixJ` from the [yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf).

eth/config.go
```diff
@@ -51,7 +51,7 @@ var DefaultLightGPOConfig = gasprice.Config{
 var DefaultConfig = Config{
 	SyncMode: downloader.FastSync,
 	Ethash: ethash.Config{
-		CacheDir:         "ethash",
+		CacheDir:         "etchash",
```
meowsbits: This config change causes the Ethash/Etchash default directory to _always_ be `~/.etchash` and `<datadir>/etchash`. Even for ETH chains, et al. This would mean that clients using core-geth on ETH would not only need to regenerate their DAG needlessly, but it would also break their filesystem API, since they'll expect that dir to be in `~/.ethash`.

We should only use a new directory default (and DAG!) for the configurations for which it's necessary.

cmd/geth/misccmd.go
```diff
@@ -35,7 +35,7 @@ var (
 		Action:    utils.MigrateFlags(makecache),
 		Name:      "makecache",
 		Usage:     "Generate ethash verification cache (for testing)",
-		ArgsUsage: "<blockNum> <outputDir>",
+		ArgsUsage: "<blockNum> <epochLength> <outputDir>",
```
meowsbits: I don't love adding a required parameter to a existing commands (since this breaks an API). Maybe instead this can accept `--<chain>` flags to have chain configuration (ie ECIP1099) context, or use a flag `--epoch.length`.

iquidus: Generally speaking I would agree, however in this case, "Generate ethash verification cache **(for testing)**". I feel it's safe to make changes. This doesnt touch chains/params it only touches ethash, generating cache/dag to outputdir then exiting.

consensus/ethash/ethash.go
```diff
@@ -412,6 +416,8 @@ type Config struct {
 	PowMode          Mode
 
 	Log log.Logger `toml:"-"`
+	// ECIP-1099
+	EtchashOnBlock *uint64
```
meowsbits: With this namespacing `ECIP1099` is equivalent to `Etchash`. I would prefer `ECIP1099Block` since it's more descriptive. This is related to the problematic idea at https://github.com/etclabscore/core-geth/pull/186#discussion_r492692919. 

Since this is the only, and only foreseeable, change to the DAG, I think I'm OK with this pattern, although it isn't as generic-y as my gut would like (deep down I think I want a map of `blockNumber:epochLength`... but probably overkill... for now ;)).

consensus/ethash/algorithm.go
iquidus: I would lean towards `epochLengthDefault` as its more descriptive at first glance. AppendixJ, altho from the paper, is a little obscure.

cmd/geth/misccmd.go
iquidus: But yeah optional and defaulting to `epochLengthDefault` makes sense. Set via --epoch.length flag.

meowsbits: _EDIT:_ no see um at your comment above, responded blind to first comment

I hear ya... but an API for testing is still an API... (and since it's useable and observable, it may be used for anything in the wild.. like miners experimenting with distributed DAGs or operators using it to generate a shared (or redundant) resource ahead of time for a pool using a newfangled microservices infrastructure..)

Is there a way we can make it an optional parameter, while maintaining established default behavior?

consensus/ethash/algorithm.go
meowsbits: Also the Yellow Paper gets edited from time to time, so Appendix J may not be an ultimate reference either.

cmd/geth/misccmd.go
iquidus: The complexity here is everywhere else uses ECIP1099FBlock defined in chainconfig, but makecache and makedag, are completely unaware of blocks or chainconfigs. So it needs to somehow be passed the param. Would an `--epoch.length` flag override epochLengthDefault/ECIP1099FBlock in the rest of the chain/ethash logic? or does it only apply to `geth makecache` and `geth makedag`?

EDIT: do we let the user ```geth --classic --epoch.length 12000``` knowing they wont be able to sync?

EDIT2: personally, im still inclined to leave as is, both commands usage state they are for testing, and the full descriptions include. 

```
This command exists to support the system testing project.
Regular users do not need to execute it.
```

My concern is a flag may cause more confusion or harm than good. e.g do i need to have it set to support the upcoming fork? (of course not, but i could see how some may think this), and if applied in general, will allow userss to break their ability to sync. On the otherside (please correct me if I'm wrong), a flag that only applies to geth makecache and geth makedag, and not ethash in general, would be new, not conforming to existing behaviour.

meowsbits: > Would an --epoch.length flag override epochLengthDefault/ECIP1099FBlock in the rest of the chain/ethash logic? or does it only apply to geth makecache and geth makedag?

My thinking was that the flag would only apply to these `makedag` and `makecache` commands (and not show up at all in `geth --help`)...

cmd/utils/flags.go
```diff
@@ -321,6 +321,11 @@ var (
 		Name:  "ethash.dagslockmmap",
 		Usage: "Lock memory maps for recent ethash mining DAGs",
 	}
+	EthashEpochLengthFlag = cli.Int64Flag{
+		Name:  "epoch.length",
+		Usage: "Sets epoch length for makecache & makedag commands",
+		Value: 30000, // TODO - iquidus
```
iquidus: Do we want to add this value to default chain/ethash configs, or can we leave as is? (it's only used by makecache and makedag)

meowsbits: Worth maybe exporting from the `ethash` package? 

Meh. I have no strong opinions, since it's pretty much set in stone and... just an adhoc constant for miscellaneous commands...

cmd/utils/flags.go
```diff
@@ -1945,16 +1947,40 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readOnly bool) (chain *core.B
...
+					localappdata := os.Getenv("LOCALAPPDATA")
+					if localappdata != "" {
+						datasetDir = filepath.Join(localappdata, "Etchash")
+					} else {
+						datasetDir = filepath.Join(home, "AppData", "Local", "Etchash")
+					}
+				} else {
+					datasetDir = filepath.Join(home, ".etchash")
+				}
+			}
```
iquidus: Here I left cache dir ethash as it exists within the network scope anyway (~/.ethereum/classic/geth/ethash).
In terms of the datasetDir (where DAGs are kept for mining), i keep flipflopping on this, the idea is so miners that mine both ethash chains and etc, dont have conflicting DAGs, however a custom path can be set with a flag as is anyway in that situation. I'm not sure whether to leave this or take it out. Thoughts?

EDIT: for clarification. default for dags is in user dir ~/.ethash outside of any network prefix.

EDIT2: also later commits change/fix the highlighted code slightly.

meowsbits: I _believe_ the subcommands that use `MakeChain` will "inherit" the value of a flag `--ethash.dagdir` setting `eth.DefaultConfig.Ethash.DatasetDir` via `utils.SetEthConfig`. If so, I think we should keep the value that's set via that flag. If that flag isn't set (ie if `eth.DefaultConfig.Ethash.DatasetDir` hasn't been overidden by the user), then go ahead and use the different-namespaced default dir if the chain config hints toward that with a non-nil `ecip1099Block`. 

This would seem to allow the following to work as expected, I think.

```
> ./build/bin/geth --classic --ethash.dagdir=/SAMSUNG/ethereumclassic/dag export blah.rlp
```

```
> ./build/bin/geth --ethash.dagdir=/SAMSUNG/ethereum/dag export blah.rlp
```

```
> ./build/bin/geth --classic export blah.rlp
```

Does that make sense?

cmd/geth/main.go
```diff
@@ -81,6 +81,7 @@ var (
 		utils.EthashDatasetsInMemoryFlag,
 		utils.EthashDatasetsOnDiskFlag,
 		utils.EthashDatasetsLockMmapFlag,
+		utils.EthashEpochLengthFlag,
```
meowsbits: Does this flag really want to be included here? If the flag is only for the `misc` commands?

cmd/utils/flags.go
```diff
@@ -1414,13 +1421,41 @@ func setTxPool(ctx *cli.Context, cfg *core.TxPoolConfig) {
...
 }
 
+func setEthashDatasetDir(ctx *cli.Context, cfg *eth.Config) {
+	switch {
+	case ctx.GlobalIsSet(EthashDatasetDirFlag.Name):
+		cfg.Ethash.DatasetDir = ctx.GlobalString(EthashDatasetDirFlag.Name)
+
+	case (ctx.GlobalBool(ClassicFlag.Name) || ctx.GlobalBool(MordorFlag.Name)) && cfg.Ethash.DatasetDir == eth.DefaultConfig.Ethash.DatasetDir:
+		// ECIP-1099 is set, use etchash dir for DAGs instead
+		home := os.Getenv("HOME")
```
meowsbits, line 1431: This copy-pasta will be able to be replaced with an exported `HomeDir` function, coming from https://github.com/etclabscore/core-geth/pull/191/files#diff-fdd625b1499a66d8f93635113abdd5e0R202.

:tada:

meowsbits:
The results of a  benchmark comparison on the `ethash` functions, comparing `master` against this branch. 

```
> benchstat .ia/master.out .ia/ecip1099.out 
name                              old time/op  new time/op  delta
CacheGeneration-12                 1.24s ± 0%   1.30s ± 0%   ~     (p=1.000 n=1+1)
SmallDatasetGeneration-12          172ms ± 0%   163ms ± 0%   ~     (p=1.000 n=1+1)
HashimotoLight-12                 10.8ms ± 0%   5.2ms ± 0%   ~     (p=1.000 n=1+1)
HashimotoFullSmall-12              106µs ± 0%    91µs ± 0%   ~     (p=1.000 n=1+1)
HashimotoFullMmap/WithLock-12     12.3µs ± 0%  13.4µs ± 0%   ~     (p=1.000 n=1+1)
HashimotoFullMmap/WithoutLock-12  12.4µs ± 0%  12.3µs ± 0%   ~     (p=1.000 n=1+1)
```

cmd/utils/flags.go
iquidus: Leaving as is for now.

meowsbits (approved):

meowsbits:
Need to fix the fork tests

PR #195, opened 2020-10-01
propose mess mainnet block

q9f:
`11_380_000`


q9f:
as per https://github.com/ethereumclassic/ECIPs/pull/388

meowsbits (approved):

PR #201, opened 2020-10-05
params: update bootnodes

q9f:


iquidus:
lgtm

meowsbits (approved):
LGTM. Just a reminder that discovery by DNS is used by default, so these will only be used if that's disabled.

q9f:
right, I forgot about that.

PR #203, opened 2020-10-07
rpc: add `txpool_removeTransaction` method

bulgakovk:
Sometimes other peers won't consider local pending transactions as valid for some reason (had a bunch of them after recent ETC 51% attacks) so they are stuck. This PR adds an RPC method for removing them.


meowsbits:
Thanks for the PR! 

OpenEthereum (aka Parity) also supports this method: https://github.com/openethereum/wiki/blob/master/JSONRPC-parity-module.md#parity_removetransaction. 

Instead of returning a boolean if the transaction was in the txpool, they return the transaction that is removed or null. If we were to adopt this, I'd suggest using this pattern instead because
- it keeps the APIs more congruent
- it provides equivalent, but more detailed, information to the user

See also: https://github.com/ethereum/go-ethereum/issues/18389.

Some reasons that other peers aren't considering local transactions valid might include being underpriced or having wrong nonces, but it's hard to know for sure.

meowsbits:
@bulgakovk I have some changes with minor formatting fixes and support of the above [commented](https://github.com/etclabscore/core-geth/pull/203#issuecomment-705598810) change.  Is it OK with you if I push to your fork? (Asking because it's on `master` and I don't want to mess anything up for you.) Changes are pushed here for review: https://github.com/etclabscore/core-geth/tree/bulgakovk-master

bulgakovk:
@meowsbits Hey your branch LGTM. Sure, feel free to push it to my fork

meowsbits:
Before merging, I'd like to assert or walk through the namespacing of the method. Removing transactions from the pool is clearly a sensitive action, and we need to make sure that that sensitivity is well reflected in the privacy of the RPC call. 

- Using `txpool` places the method in a shared space with public methods like `txpool_status` and `txpool_content`.
- Parity uses `parity_removeTransaction`.

The call should be available over IPC by default, but not available via HTTP or WS by default.
The call should __not__ be made available at the same time/in the same way as the other existing `txpool_` methods. This is my primary concern with the existing implementation.

bulgakovk:
@meowsbits Does `admin_*` namespace look better? If so we can call `admin_removePendingTransaction`

meowsbits:
Yes, I think that is probably the safest option.

meowsbits:
Or `debug_removePendingTransaction`

bulgakovk:
>The call should not be made available at the same time/in the same way as the other existing txpool_ methods. 

Could you please elaborate on this? Do you mean we need a separate method in `TxPool` for the PRC method?

>The call should be available over IPC by default, but not available via HTTP or WS by default.

Can you please point me to an example of this configuration?

meowsbits:
> Could you please elaborate on this?

I meant that `txpool_removeTransaction` should have privacy setting distinct from the existing `txpool_` namespace. So given `--http.api=txpool`, the method `txpool_removeTransaction` should NOT be available over HTTP along with the safe methods like `txpool_status`. This issue is addressed by using the `admin_` or `debug_` namespaces, which are intended to contain sensitive methods.

> Can you please point me to an example of this configuration?

`admin_` and `debug_` methods are examples of this, actually :)

meowsbits:
@bulgakovk You said this issue happened/happens for you "sometimes." Was the issue mostly restricted to the anomalous large reorg events?

I'm asking because if the issue is rare, then I would want to give pause to adding a dedicated RPC method to address it since it can be resolved in other ways, eg. restarting geth and `rm <datadir>/geth/transactions.rlp`.

bulgakovk:
> @bulgakovk You said this issue happened/happens for you "sometimes." Was the issue mostly restricted to the anomalous large reorg events?

Yes, it was restricted to the recent 51% attacks.


> I'm asking because if the issue is rare, then I would want to give pause to adding a dedicated RPC method to address it since it can be resolved in other ways, eg. restarting geth and `rm <datadir>/geth/transactions.rlp`.

This is how we solve that issue currently and this is very annoying. Stopping `geth` means you also have to load balance traffic and requires too many manual steps. I think ideally the issue should be handled programatically and a new RPC method sounds like a good fit for that. Does it sound reasonably to you? If so I'll be happy to finish fixing the security-related issues.

meowsbits:
Thanks for clarifying. Yes, I think it's OK to add the new RPC method as long as the security issues are handled, thank you!

bulgakovk:
So after [this](https://github.com/etclabscore/core-geth/pull/203/commits/6705d3a210f8147908fff51feab063e38104f823) commit I started `geth` with the following parameters:
```
--datadir
<>
--rpc
--rpcaddr
0.0.0.0
--rpcport
8545
--rpcapi
"admin,eth,net,personal,txpool,web3"
--syncmode=full
--gcmode=archive
--goerli
```
so `debug` is missing in the `--rpcapi` parameter. Calling
```
curl --location --request POST 'http://localhost:8545' \
--header 'Content-Type: application/json' \
--data-raw '{
    "id": 1,
    "method": "debug_removePendingTransaction",
    "params": ["0x92cb8ac282ca070642dbb03744c2cbc1ac9c2662252b581d6bb1a8e14584df14"]
}'
```
produces this response
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32601,
        "message": "the method debug_removePendingTransaction does not exist/is not available"
    }
}
```

Does it solve the security concern you had?

internal/web3ext/web3ext.go
```diff
@@ -766,7 +766,7 @@ web3._extend({
 	methods: [
 		new web3._extend.Method({
 			name: 'removeTransaction',
-			call: 'txpool_removeTransaction',
+			call: 'debug_removeTransaction',
```
meowsbits: This method declaration needs to move to `DebugJs`.

bulgakovk: Looks like GitHub didn't index the changes properly. The declaration was transplanted to the `DebugJs` definition already: https://github.com/etclabscore/core-geth/pull/203/files#diff-743816a1d619b592f395ca0641e2f96df78708f35726f1099da4696a1a9384a0R476

bulgakovk:
Hi! Do you have any updates on this? :)

meowsbits (approved):

meowsbits:
:heavy_check_mark: Done! Thank you @bulgakovk!

bulgakovk:
Great, thank you! So the feature will be available once the next `geth` release is merged into this repository, right?

PR #210, opened 2020-10-14
Merge/foundation master 20201014

ziogaschr:


q9f:
I would prefer if we pull releases instead of master. Is that possible?

ziogaschr:
Oh good one. I will do that. Is it ok to leave it for later today? @q9f

meowsbits:
@q9f I definitely understand the preference for generally pulling from tags instead of master. With that, I think it's worth some conversation because there's some nuances either way, in my opinion.

- ethereum/go-ethereum release and master branches are sometimes quite different, where they'll do significant backports to transfer features or bug fixes (this isn't directly applicable to an argument either way, just contextual)
- although I think that probably the release versions at ethereum/go-ethereum are generally _somewhat_ more stable than master, I'm not convinced that they're _largely_ more stable. 
- proposed change sets like this one are merging master (upstream) to master (local)
- smaller and more frequent upstream merges make review (for the merger and the reviewer) easier, more incremental

In general, I think it makes sense to
- Review each incoming upstream commit individually, and to understand them as a set. We need to understand what we're pulling in; what's sensitive, and what's mundane.
- Generally aim to tag local releases such that they include merges with upstream through some respective tag. That is, to not tag local releases that include merges from upstream which are "uncapped" by upstream tags.
- The point above, however, doesn't depend on the actual upstream-to-master change sets being directly derivative of upstream tags. 

I'm not trying to definitively argue one way or the other, just to provide some context and counterargument for sake of opening a conversation about the protocol and practice.

Brainstorming, one idea might be to use an upstream tracking branch (local master and upstream master hybrid) so that we can keep up a practice of frequent small chunks of upstream merges, while not being forced to tag releases on unfinished (or un-"capped") work. This hybrid branch would have all conflicts readily resolved versus local master, and thus would be trivial to merge to master prior to a local release.

ziogaschr:
I used the same PR to merge with the next release/tag (adds one more commit). I haven't noticed the maintain the release branches, so I used the tag to merge this time.
Let me know if you want to clean it up (as we have 2 merges in the history now) and create a new PR.

q9f:
I'm actually ok with merging upstream master into our master but then I would recommend doing releases on release branches instead.

There are many ways to tackle this and I didn't mean to disrupt your process here. Happy to have a quick call this week talking through different strategies.

meowsbits:
Roger that. And thanks, @ziogaschr! I'm happy with the history as-is.

meowsbits (changes requested):
- [x] set DNS defaults for `cmd/faucet` for non-foundation networks, rel d9890a6

... and that's all I found!

meowsbits (approved):

meowsbits:
@ziogaschr PTAL last two commits, rel https://github.com/etclabscore/core-geth/pull/210#pullrequestreview-516707185. Requesting your review ;)

ziogaschr:
@meowsbits loool, I was checking too, hasn't thought you will be so kind to provide a fix. Checking it..

PR #211, opened 2020-10-15
Support trace API similar to OpenEthereum/Parity's one

### LINKED ISSUE #100: Question: I am a user from exchange, we want migrate from parity to core-get
I am a user from exchange, we want migrate from parity to core-geth
so what's the difference about rpc access between parity and core-geth

ziogaschr:
This PR adds a compatible trace API with [OpenEthereum's one](https://openethereum.github.io/wiki/JSONRPC-trace-module#trace_block). 

At the moment we have implemented:

- [x] trace_transaction
- [x] trace_block

For a node to enable this feature, it has to start with `--vmdebug` flag.

For the best performance node has to be synced in **archive** mode.
In console the `trace` API can be used.
Last, the `trace` RPC API is available, though it have to be enabled.

Fixes  [#100](https://github.com/etclabscore/core-geth/issues/100).

---

**Work in Progress. Opened a PR to handle discussion over here.**


eth/tracers/tracer.go
```diff
@@ -541,6 +541,13 @@ func (jst *Tracer) CaptureStart(from common.Address, to common.Address, create b
 	return nil
 }
 
+func (jst *Tracer) CaptureExtraContext(inputs map[string]interface{}) error {
```
ziogaschr, line 569: @iquidus @meowsbits I am thinking to make the `map[string]interface{}` a Struct and parse it in here with reflect. Do you think it's ok for now like it is? Is it overkill to use reflect for this?

meowsbits: For now I think it would be overkill... although worth considering for a refactor later on. The `Tracer.ctx` field is used in several ways that seem to depend on the type-agnosticism of the `map[string]interface{}`, eg. `txTraceTask.extraContext`, so this may be something of a rabbit hole.

ziogaschr: That's true, originally `Tracer.ctx` holds transaction context. I have added `CaptureExtraContext` in order to be able to extend context with block related stuff. For this reason I was thinking to make only the `inputs` in `CaptureExtraContext` a struct. Though at the end, they will be added down to the `Tracer.ctx` being a `map[string]interface{}`

eth/api_tracer_oe.go
```diff
@@ -155,9 +155,7 @@ func (api *PrivateTraceAPI) Block(ctx context.Context, number rpc.BlockNumber, c
 		if err := json.Unmarshal(result.Result.(json.RawMessage), &tmp); err != nil {
 			return nil, err
 		}
-		for _, item := range tmp {
-			results = append(results, item)
-		}
+		results = append(results, tmp...)
```
ziogaschr: I haven't run actual perf tests, but I've read somewhere that this might be much slower on bigger lists (which is not the case here).

ziogaschr:
@meowsbits thanks for your input and commits

eth/api_tracer_oe.go
meowsbits: Interesting! I just wrote some quick benchmarks to investigate, included in https://github.com/etclabscore/core-geth/pull/211/commits/69d7e09ae8b6d41889de30d328b48ffdecbe4469.

```go
// BenchmarkTraceResultsAppend1 compares performance against BenchmarkTraceResultsAppend2,
// comparing the performance of different ways of appending items to slices.
// This is used in PrivateTraceAPI#Block appending results to the traceResults value.
func BenchmarkTraceResultsAppend1(b *testing.B) {
	traceResults := []interface{}{}
	for i := 0; i < 10000; i++ {
		traceResults = append(traceResults, i)
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		// results := make([]interface{}, len(traceResults))
		results := []interface{}{}
		for _, it := range traceResults {
			results = append(results, it)
		}
	}
}

// BenchmarkTraceResultsAppend2 (see comment for BenchmarkTraceResultsAppend1).
func BenchmarkTraceResultsAppend2(b *testing.B) {
	traceResults := []interface{}{}
	for i := 0; i < 10000; i++ {
		traceResults = append(traceResults, i)
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		// results := make([]interface{}, len(traceResults))
		results := []interface{}{}
		results = append(results, traceResults...)
	}
}
```

Here are the results:

```txt
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/eth
BenchmarkTraceResultsAppend1
BenchmarkTraceResultsAppend1-12    	    3742	    288702 ns/op
BenchmarkTraceResultsAppend2
BenchmarkTraceResultsAppend2-12    	   20686	     59821 ns/op
PASS
```

ziogaschr: I love you man. Code says always the truth.

eth/api_tracer.go
```diff
@@ -352,23 +353,29 @@ func (api *PrivateDebugAPI) traceChain(ctx context.Context, start, end *types.Bl
 
 // TraceBlockByNumber returns the structured logs created during the execution of
 // EVM and returns them as a JSON object.
-func (api *PrivateDebugAPI) TraceBlockByNumber(ctx context.Context, number rpc.BlockNumber, config *TraceConfig) ([]*txTraceResult, error) {
+func traceBlockByNumber(ctx context.Context, eth *Ethereum, number rpc.BlockNumber, config *TraceConfig) ([]*txTraceResult, error) {
```
meowsbits, line 356: What's the reasoning behind pulling the receiver out and using the `eth *Ethereum` arg instead? (Instead of just making the method private?) Not against it, just curious.

ziogaschr: Just for using it in both **PrivateDebugAPI** and **PrivateTraceAPI** [here](https://github.com/etclabscore/core-geth/pull/211/files#diff-e98ceee1b6a777b2ee45a6f6527b9d51da990c691cbb19997c8c104ca9dd4725R136). I would be interested if you have a better idea on doing this, as I don't like it much. I did the same for more functions.

eth/api_tracer_oe.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/ethereum/go-ethereum/consensus/ethash"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/internal/ethapi"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+// OpenEthereumTrace A trace in the desired format (Parity/OpenEtherum) See: https://openethereum.github.io/wiki/JSONRPC-trace-module
+type OpenEthereumTrace struct {
+	Action              TraceRewardAction `json:"action"`
+	BlockHash           *common.Hash      `json:"blockHash"`
```
meowsbits: Does this need to be a pointer? Since `BlockNumber` isn't, and neither type uses `omitempty`, it seems there is an assumption that the block context will always be available. Or?

eth/api_tracer_oe.go
```diff
@@ -0,0 +1,181 @@
...
+// The correct name will be TraceBlockByNumber, though we want to be compatible with OpenEthereum trace module.
+func (api *PrivateTraceAPI) Block(ctx context.Context, number rpc.BlockNumber, config *TraceConfig) ([]interface{}, error) {
+	// Fetch the block that we want to trace
+	var block *types.Block
+
+	switch number {
+	case rpc.PendingBlockNumber:
+		block = api.eth.miner.PendingBlock()
+	case rpc.LatestBlockNumber:
+		block = api.eth.blockchain.CurrentBlock()
```
meowsbits: Should we handle `case rpc.EarliestBlockNumber` here too? 

EDIT: Actually the method and apparently some of associated "helper" functions don't support genesis block tracing. Which makes sense, I guess, but seems kind of silently assumed... ?

eth/api_tracer_oe.go
ziogaschr: No it doesn't. Nice catch. It actually started wrong from wrong assumptions from OpenEthereum trace outputs and copying them from turbo-geth as starting structs.

eth/api_tracer_oe.go
ziogaschr: Probably not, I copied this from the `debug_traceBlock` where it's neither being handled. The reason it's not being handled is, that we can't retrieve block trace for the genesis block. We can add it if a reason exists, though the current logic fails when it tries to read the state of the parent block, as genesis has no parent block.

p.s.: I have extracted the logic of getting a block from a block number in a stash I have for `trace_filter` support.

eth/api_tracer_oe.go
```diff
@@ -50,7 +50,7 @@ type TraceRewardAction struct {
 	RewardType string          `json:"rewardType,omitempty"`
 }
 
-// setConfigTracerToOpenEthereum forces the Tracer to the OpenEthereum one
+// setConfigTracerToOpenEthereum forces the Tracer to the OpenEthereum one, mutating the argument value.
 func setConfigTracerToOpenEthereum(config *TraceConfig) *TraceConfig {
 	if config == nil {
```
ziogaschr: @meowsbits when `config == nil`, there is no actual pointer to assign to. In this case the config is being scoped in `setConfigTracerToOpenEthereum`. Assigning to an indirect pointer, can't work in this case either. If we want to do it without assigning from the caller func, then we have to call it like:

```go
if config == nil {
	config = new(TraceConfig)
}
setConfigTracerToOpenEthereum(config)
```

eth/api_tracer_oe.go
ziogaschr: @meowsbits if we want to keep 100% compatibility with OpenEthereum tracer, then they return the below for rewards, as they use the same struct for both `trace_block` and `trace_transaction`.

```
    {
            "action": {
                "author": "0x877bd459c9b7d8576b44e59e09d076c25946f443",
                "rewardType": "block",
                "value": "0x4563918244f40000"
            },
            "blockHash": "0x8df024322173d225a09681d35edeaa528aca60743a11a70f854c158862bf5282",
            "blockNumber": 553416,
            "result": null,
            "subtraces": 0,
            "traceAddress": [],
            "transactionHash": null,
            "transactionPosition": null,
            "type": "reward"
        }
```

In our case where we use it only for block rewards, it's totally fine to remove the `TransactionHash` and `TransactionPosition`. What do you think? Shall we keep them?

q9f:
Add the `Y1` label because I believe this should be discussed further upstream once we are satisfied here.

eth/api_tracer_oe.go
ziogaschr: @meowsbits what do you think on this? do you have any better ideas?

ziogaschr: Reverted for now, so as it works on external review. Will leave this review unresolved though.

ziogaschr:
At this point we reached 100% compatibility with OpenEthereum tracers for `trace_block` and `trace_transaction` at Mordor testnet chain.

meowsbits:
State tests are failing, looks like it could be an issue with CREATE gas. Possibly related to c32bcd6abab903642b5f278fe0e68b832c2a5756. 

To re/produce: `go test ./tests/`

They look mostly like `out of gas` cases, maybe?

consensus/ethash/consensus.go
```diff
@@ -587,28 +587,43 @@ var (
 // AccumulateRewards credits the coinbase of the given block with the mining
 // reward. The total reward consists of the static block reward and rewards for
 // included uncles. The coinbase of each uncle block is also rewarded.
-func accumulateRewards(config ctypes.ChainConfigurator, state *state.StateDB, header *types.Header, uncles []*types.Header) {
+// func AccumulateRewards(config ctypes.ChainConfigurator, state *state.StateDB, header *types.Header, uncles []*types.Header) {
```
ziogaschr: @meowsbits @iquidus can you review this function and the ones that being used inside for any logical issues?

consensus/ethash/consensus.go
```diff
@@ -587,28 +587,43 @@ var (
 // AccumulateRewards credits the coinbase of the given block with the mining
```
iquidus: This description no longer matches the function.

consensus/ethash/consensus.go
```diff
@@ -587,28 +587,43 @@ var (
...
 // reward. The total reward consists of the static block reward and rewards for
 // included uncles. The coinbase of each uncle block is also rewarded.
-func accumulateRewards(config ctypes.ChainConfigurator, state *state.StateDB, header *types.Header, uncles []*types.Header) {
+// func AccumulateRewards(config ctypes.ChainConfigurator, state *state.StateDB, header *types.Header, uncles []*types.Header) {
+func AccumulateRewards(config ctypes.ChainConfigurator, header *types.Header, uncles []*types.Header) (*big.Int, []*big.Int) {
```
iquidus: Perhaps rename func to `CalcRewards`? Or for consistency with reward calculation func's `GetRewards`

core/vm/logger.go
```diff
@@ -106,7 +106,7 @@ func (s *StructLog) ErrorString() string {
 // if you need to retain them beyond the current call.
 type Tracer interface {
 	CaptureStart(from common.Address, to common.Address, create bool, input []byte, gas uint64, value *big.Int) error
-	CaptureState(env *EVM, pc uint64, op OpCode, gas, cost uint64, memory *Memory, stack *Stack, rStack *ReturnStack, rData []byte, contract *Contract, depth int, err error) error
+	CaptureState(env *EVM, pc uint64, op OpCode, gas, availableGas, cost uint64, memory *Memory, stack *Stack, rStack *ReturnStack, rData []byte, contract *Contract, depth int, err error) error
```
meowsbits: @ziogaschr I think you mentioned yesterday that this parameter may be removable since this value is equivalent to `env.callGasTemp`. If so, is it worth doing?

ziogaschr: At this point removed the extra `availableGasCopy` and use the `in.evm.callGasTemp`. This one can be removed if I make it public as `in.evm.CallGasTemp`. What do you think about it?

consensus/ethash/consensus.go
```diff
@@ -584,31 +584,46 @@ var (
...
+
+	return reward, uncleRewards
+}
+
+// accumulateRewards credits the coinbase of the given block with the mining
+// reward. The coinbase of each uncle block is also rewarded.
+func accumulateRewards(config ctypes.ChainConfigurator, state *state.StateDB, header *types.Header, uncles []*types.Header) {
+	minerReward, uncleRewards := GetRewards(config, header, uncles)
+	for i, uncle := range uncles {
+		if i < len(uncleRewards) {
```
meowsbits: This is an unnecessary condition since each uncle will always get one reward (and they're in a known and constant order), so `len(uncleRewards)` will always be `len(uncles)`.

consensus/ethash/consensus.go
meowsbits: Have done. LGTM, and the passing tests indicate positive too.

meowsbits (dismissed):
:x: `go test ./eth/tracers/...` is failing

Eg. https://travis-ci.org/github/etclabscore/core-geth/jobs/741433701#L960

core/vm/logger.go
meowsbits: I think it's OK to make it public because that follows the pattern that we've already "allowed" with the introduction of `EVM.CallErrorTemp`, and IMO publicizing a field there is less "expensive" than adding a parameter to a shared interface.

ziogaschr:
Nice catch @meowsbits. I fixed the existing foundation `call_tracer` which I broke with the changes with regards gas.

p.s.: Haven't checked the Travis failures as I noted them as broken most of the times a while back. Have we fixed them?

meowsbits:
I'm going to merge this, but just noting the following CI suites failed at the same test, which is unusual. I could, however, not 
reproduce this locally so I'm assuming this as unrelated issue, though one that we may need to follow up with.

This is a core-geth -specific integration test. 

```
--- FAIL: TestGenerateBlockAndImport_CG1 (0.00s)

    --- FAIL: TestGenerateBlockAndImport_CG1/all-ethash (3.01s)

        worker_cg_test.go:77: timeout: 1

FAIL
```

- https://travis-ci.org/github/etclabscore/core-geth/jobs/741651011#L990
- https://travis-ci.org/github/etclabscore/core-geth/jobs/741651027#L1188

meowsbits (approved):

PR #212, opened 2020-10-15
1099 miner DAG pre-generation fix

iquidus:
Fixes an issue with pre-generation in which a miner would pre-generate `epoch + 1` always. This is correct behaviour with the exception of the activation of 1099, in which the miner should pre-generate `(epoch+1) / 2`.

This PR adds additional logic to properly handle the activation of ecip-1099.


meowsbits:
Is there an existing test for this issue? Is it possible to add one? Desirable to?

meowsbits (approved):

PR #137, opened 2020-07-06
rpc: rpc_discover by reflection

meowsbits:
Replaces #90. There's a lot of good, detailed information and storied history there. Check it out. However, that branch got stale waiting for some upstream dependency stuff, so this is a new, much cleaner and master-happy, version. 

The change set:
- [x] Removes the existing OpenRPC `rpc.discover` endpoint pattern. That implementation depended on a hard-coded static document, located in `internal/openrpc` and depended on custom logic in the `rpc` package to serve it.  This clears all that out.
- Establishes dynamically-generated service discovery information for IPC, HTTP, and WS node services. 

I run the following script to generate and validate the document. If you want to do something like this, you'll just need to tweak a couple of personal hardcoded values and make sure you have the dependencies.

```sh
> cat .ia/dev-openrpc2.sh
#!/usr/bin/env bash
set -e

main() {
        make geth
        pkill -f ./build/bin/geth || true

        ./build/bin/geth --port 30313 --datadir=/tmp/gethddd --nodiscover --maxpeers=0 --rpc --rpcapi=admin,debug,eth,ethash,miner,net,personal,rpc,txpool,web3 >/tmp/geth.log 2>&1 &
        disown
        gethpid=$!

        echo "Gethpid: ${gethpid}"

        rm -fr /tmp/gethddd
        onexit() {
                kill -9 ${gethpid}
                pkill ./build/bin/geth
                rm -fr /tmp/gethddd
        }
        trap onexit EXIT

        # Wait for geth to start up.
        echo "Sleeping 3"
        sleep 3

        # Post rpc.discover request.

        gethroot="$HOME/go/src/github.com/ethereum/go-ethereum"
        http --json POST http://localhost:8545 id:=$(date +%s) method='rpc_discover' params:='[]' | jj -p 'result' | tee "${gethroot}/spec.json"

        # Validate response.

        validatorroot="$HOME/dev/etclabscore/openrpc-validator-cli"
        node "$validatorroot/build/index.js" "$gethroot/spec.json" || { cat /tmp/geth.log && echo "FAILED" && exit 1; }

        kill ${gethpid}

        # Publish gist if OK and not flagged off.
        if [[ $1 == nogist ]]; then
                echo Skipping gist update
        else
                echo "Updating gist 4da4c08765679dac1899543002d1f545 from ${gethroot}/spec.json"
                gist -u 4da4c08765679dac1899543002d1f545 "${gethroot}/spec.json" >/dev/null 2>&1 &
        fi

[...]

PR #220, opened 2020-10-30
Feat/openrpc discover 20201024

meowsbits:
Supersedes #137 

I update a representative OpenRPC document generated by this implementation (via development tests) at https://gist.github.com/meowsbits/4da4c08765679dac1899543002d1f545.  This document is rendered by the OpenRPC playground at https://playground.open-rpc.org/?url=https://gist.githubusercontent.com/meowsbits/4da4c08765679dac1899543002d1f545/raw/spec.json.


BelfordZ:
almost there!

Is there anyway to configure `trimpath` to ignore the OpenRPC doc?

meowsbits:
The `-trimpath` pain won't be solved by just ignoring the OpenRPC doc, since the problem it causes is that it breaks filepaths for AST traversal in general. 

So far, my thinking is that it can be worked around by creating basically an environment variable that etclabscore/go-openrpc-reflect would fall back to in case its AST steps error with "file not found." I'm not sure if there's a way to corroborate that failure at build time with a check against the `go build` flags, but would be nice.

node/node_test.go
```diff
@@ -47,6 +52,26 @@ func testNodeConfig() *Config {
 	}
 }
 
+// TestFuncForPC is a sanity-check PoC demonstrating runtime caller/frame/pc availability and functionality.
+// TODO(meowsbits) Remove this, it's debug only.
```
ziogaschr: @meowsbits do we want to remove this one?

params/vars/environment.go.go
```diff
@@ -0,0 +1,66 @@
+package vars
```
ziogaschr: @meowsbits does this file filename has `.go.go` by intention?

node/openrpc.go
```diff
@@ -0,0 +1,472 @@
...
+		outs := make([]reflect.Type, method.Func.Type().NumOut())
+		for i := 0; i < method.Func.Type().NumOut(); i++ {
+			outs[i] = method.Func.Type().Out(i)
+		}
+
+		// If an error is returned, it must be the last returned value.
+		switch {
+		case len(outs) > 2:
+			return false
+		case len(outs) == 1 && isErrorType(outs[0]):
```
ziogaschr: Is there an edge case that this case covers? In this case we might want to add a comment reasoning on it.

Can it be that we had in mind the opposite `!isErrorType(outs[0])`, though it doesn't make much sense as there are API methods like `web3_sha3` that return a non error output.

node/node_test.go
meowsbits: Yeap, nice catch. Nothing special here... just developer scraps.

params/vars/environment.go.go
meowsbits: Nope, accidental.

node/openrpc.go
meowsbits: Yea, it would be helpful to include examples in comments here. You can cross reference the `rpc` package to check corrrectness... in `rpc/server.go` I believe (?). The idea (as from the comment on #L134) is that `If an error is returned, it must be the last returned value`, which is just an enforced convention for the signature pattern for eligible methods.

- OK: `func a() error`
- OK: `func b() (int, error)`
- OK: `func c() int`
- OK: `func d()`
- NOTOK: `func e() (error, int)`

node/node_test.go
ziogaschr: Resolved by https://github.com/etclabscore/core-geth/pull/220/commits/6cf6a48b2c34e40e56992fdbfdbf60adc0a2853b

params/vars/environment.go.go
ziogaschr: Fixed by https://github.com/etclabscore/core-geth/pull/220/commits/0f2dcd42208d65424cf9a84137ce3cad2f94f19e

node/openrpc.go
ziogaschr: Resolved by https://github.com/etclabscore/core-geth/pull/220/commits/cfd5f28e7559180615b96c54a203ab144812f928

ziogaschr: Even better https://github.com/etclabscore/core-geth/pull/220/commits/690b2d392e2e57f49bc2403d93622f8b16440726. Thanks @meowsbits

ziogaschr (approved):

PR #223, opened 2020-11-06
Bit boundary fix for the DAG generation routine

slavikus:
If DAG size is exceeding the maximum 32 bit unsigned value, the DAG generation will produce incorrect results depending on the number of threads available on the host machine. Switch the generation to 64 bit variable to fix this.

Note that a hard fork may be required on the ETC network for the time being as we're already on the epoch 385 which generates a DAG file of 4.008 Gb, causing some nodes to reject valid blocks, depending on the conditions under which they have generated a new DAG file under.

### DIFF
--- a/consensus/ethash/algorithm.go
+++ b/consensus/ethash/algorithm.go
@@ -320,16 +320,16 @@ func generateDataset(dest []uint32, epoch uint64, epochLength uint64, cache []ui
 			keccak512 := makeHasher(sha3.NewLegacyKeccak512())
 
 			// Calculate the data segment this thread should generate
-			batch := uint32((size + hashBytes*uint64(threads) - 1) / (hashBytes * uint64(threads)))
-			first := uint32(id) * batch
+			batch := uint64((size + hashBytes*uint64(threads) - 1) / (hashBytes * uint64(threads)))
+			first := uint64(id) * batch
 			limit := first + batch
-			if limit > uint32(size/hashBytes) {
-				limit = uint32(size / hashBytes)
+			if limit > uint64(size/hashBytes) {
+				limit = uint64(size / hashBytes)
 			}
 			// Calculate the dataset segment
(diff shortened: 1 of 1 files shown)


meowsbits:
Thanks for this. We are running a few sync tests and general otherwise sanity checks, but in general this looks good and unless we find something unexpected we'll have it merged very soon.

meowsbits:
I just completed a successful test where I
- `rm -rf`d the Ethash DAG datadir and cachedir
- `debug.setHead` to block 11m (which was past my previous fast sync pivot, thus enabling fast sync mode)
- resynced `--classic` from there

Node completed sync successfully. 

I have also run some basic checks on the interoperability of the ethash cache (etchash) data between the current `master` version and this PR for a non-mining node to show at least that things won't explode when users upgrade.

It will probably be safest for miners to regenerate their DAG dirs and caches when they upgrade.

meowsbits (approved):
LGTM. 

Please note comments at https://github.com/ethereum/go-ethereum/pull/21793#issuecomment-723051563, and https://github.com/ethereum/go-ethereum/pull/21793#issuecomment-723084523.

meowsbits:
@iquidus, please give your review and merge if and when OK by you too. We'll aim for tagging and publishing a release today.

iquidus (approved):

bobsummerwill:
@meowsbits "It will probably be safest for miners to regenerate their DAG dirs and caches when they upgrade."

With the https://github.com/etclabscore/core-geth/releases/tag/v1.11.17 release today, would that be your only recommendation?

For all Core-Geth users to upgrade and further for miners to regenerate?

And it does not appear that any blocks have been mined with broken DAGs, eh?   We are not dependent on replicating the bug into the future, right?

meowsbits:
Yes, yes, yes, yes! :1st_place_medal:

Issue #222, opened 2020-11-05
geth copydb command fails


meowsbits:
#### System information

```
> ./build/bin/geth version
CoreGeth
Version: 1.11.17-unstable
Git Commit: e447981171d6dd46e9f893a8f679d645ce56604b
Git Commit Date: 20201105
Architecture: amd64
Protocol Versions: [65 64 63]
Go Version: go1.15
Operating System: linux
GOPATH=/home/ia/go
GOROOT=/home/ia/go1.15.linux-amd64
```

#### Expected behaviour

`geth copydb` copies a database successfully or logs a sensible error rationalizing its refusal and exits.

#### Actual behaviour

1. Case when source db does not contain block data (no import progress). Program logs `retrieved ancestor is invalid` and exits `1`. Target database does not exist.

2. Case when source db _does_ contain block data (some import progress). Program panics. This appears to be an issue with ancient/kv relationship. It is expected that the imported historical data should have gone straight to ancient, and the KV should be essentially empty.

#### Steps to reproduce the behaviour

__Case 1__

```
> ./build/bin/geth --datadir ./dd --exec 'exit;' console

[...]

> [[ -d ./dd/geth/chaindata ]] && echo ok
ok

> ./build/bin/geth copydb ./dd/geth/chaindata /tmp/geth_dd_copy
INFO [11-05|09:59:13.397] Maximum peer count                       ETH=50 LES=0 total=50
INFO [11-05|09:59:13.397] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [11-05|09:59:13.399] Set global gas cap                       cap=25000000
INFO [11-05|09:59:13.399] Allocated cache and file handles         database=/home/ia/.ethereum/geth/chaindata cache=512.00MiB handles=524288
INFO [11-05|09:59:13.449] Opened ancient database                  database=/home/ia/.ethereum/geth/chaindata/ancient
INFO [11-05|09:59:13.454] Found stored genesis block               config="{ChainID: 1 Homestead: 1150000 DAO: 1920000 DAOSupport: true EIP150: 2463000 EIP155: 2675000 EIP158: 2675000 Byzantium: 4370000 Constantinople: 7280000 ConstantinopleFix: 7280000 Istanbul: 9069000 MuirGlacier: 9200000 Engine: ethash}"
[...]

PR #227, opened 2020-11-09
main: fix usage for copydb command

meowsbits:
The help text noted the mandatory first arg,
but not the mandatory second arg, which is
the path to the ancient data dir.

---

Rel #222.

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -178,7 +178,8 @@ The export-preimages command export hash preimages to an RLP encoded stream`,
 		},
 		Category: "BLOCKCHAIN COMMANDS",
 		Description: `
-The first argument must be the directory containing the blockchain to download from`,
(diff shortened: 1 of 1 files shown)


ziogaschr (approved):
LGTM.

I am wondering if it would make sense to fallback to the default path inside chaindata, if no 2nd argument is provided. On the other hand asking it explicitly on such command makes sense.

`--datadir.ancient value               Data directory for ancient chain segments (default = inside chaindata)`

meowsbits:
At this point I'm happy just improving the help usage. I don't mind the program having strict demands about usage, and generally am skeptical of fallback (ermm... "magic" ;) programming. 

Having just typed that and giving it 1 more minute of thought I think... I'm coming around to the fallback idea... since it exactly echoes the behavior of `geth`'s use otherwise. And probably most people using the command would need to look up where that directory was (and what the `ancient` store even was, too :joy:). If you want to do this, let's go ahead, but in a new PR though ;)

PR #232, opened 2020-11-12
Feat/go 1155

meowsbits:
Upgrades Go to v1.15.5 for artifacts builds, which includes an important security fix. 

This also includes a new flag `-dlgo` for `go run ./build/ci.go install`, which is configured to use v1.15.5 as the download target.


ziogaschr (approved):
LGTM

PR #236, opened 2020-11-13
Fixes 234

### LINKED ISSUE #234: incorrect seedhash for the eth_GetWork reply (ECIP-1099)
ethash/sealer is generating incorrect seedhash for the eth_GetWork reply. 

In the current Mordor testnet it returns 0x0518e1373ff88ccabb28493cac10cb0731313135d880dae0d846be6016ab9acf instead, which is valid for the 89th epoch and not 88th like specified in the ECIP-1099

iquidus:
Fixes an issue in which, makeWork was producing an incorrect seedHash post ecip-1099 activation, on previously odd epochs (e.g 89).

fixes #234

### DIFF
--- a/consensus/ethash/ethash.go
+++ b/consensus/ethash/ethash.go
@@ -301,8 +301,9 @@ func isBadCache(epoch uint64, epochLength uint64, data []uint32) (bool, string) 
 // generate ensures that the cache content is generated before use.
 func (c *cache) generate(dir string, limit int, lock bool, test bool) {
 	c.once.Do(func() {
-		size := cacheSize(c.epoch*c.epochLength+1, c.epoch)
-		seed := seedHash(c.epoch*c.epochLength + 1)
+		epochStartBlock := calcEpochBlock(c.epoch, c.epochLength)
+		size := cacheSize(epochStartBlock, c.epoch)
+		seed := seedHash(epochStartBlock)
 		if test {
 			size = 1024
 		}
@@ -349,7 +350,7 @@ func (c *cache) generate(dir string, limit int, lock bool, test bool) {
 		}
 		// Iterate over all previous instances and delete old ones
 		for ep := int(c.epoch) - limit; ep >= 0; ep-- {
-			seed := seedHash(uint64(ep)*c.epochLength + 1)
+			seed := seedHash(calcEpochBlock(uint64(ep), c.epochLength))
 			path := filepath.Join(dir, fmt.Sprintf("cache-R%d-%x%s", algorithmRevision, seed[:8], endian))
 			os.Remove(path)
 		}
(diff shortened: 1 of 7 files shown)


ziogaschr (approved):

meowsbits:
Rel #238

consensus/ethash/sealer.go
```diff
@@ -344,8 +344,10 @@ func (s *remoteSealer) loop() {
 //   result[3], hex encoded block number
 func (s *remoteSealer) makeWork(block *types.Block) {
 	hash := s.ethash.SealHash(block.Header())
+	epochLength := calcEpochLength(block.NumberU64(), s.ethash.config.ECIP1099Block)
+	epoch := calcEpoch(block.NumberU64(), epochLength)
 	s.currentWork[0] = hash.Hex()
-	s.currentWork[1] = common.BytesToHash(SeedHash(block.NumberU64())).Hex()
+	s.currentWork[1] = common.BytesToHash(SeedHash(epoch*epochLength + 1)).Hex()
```
meowsbits: `SeedHash` is also used in the `debug` API. 

```go
// SeedHash retrieves the seed hash of a block.
func (api *PublicDebugAPI) SeedHash(ctx context.Context, number uint64) (string, error) {
	block, _ := api.b.BlockByNumber(ctx, rpc.BlockNumber(number))
	if block == nil {
		return "", fmt.Errorf("block #%d not found", number)
	}
	return fmt.Sprintf("0x%x", ethash.SeedHash(number)), nil
}
```

> `internal/ethapi/api.go`

Does the logic there need to change too?

internal/ethapi/api.go
```diff
@@ -1856,12 +1856,13 @@ func (api *PublicDebugAPI) PrintBlock(ctx context.Context, number uint64) (strin
...
 
 // SeedHash retrieves the seed hash of a block.
-func (api *PublicDebugAPI) SeedHash(ctx context.Context, number uint64) (string, error) {
+func (api *PublicDebugAPI) SeedHash(ctx context.Context, number uint64, epochLength uint64) (string, error) {
 	block, _ := api.b.BlockByNumber(ctx, rpc.BlockNumber(number))
 	if block == nil {
 		return "", fmt.Errorf("block #%d not found", number)
 	}
-	return fmt.Sprintf("0x%x", ethash.SeedHash(number)), nil
+	epoch := number / epochLength
```
ziogaschr: This changes the API signature, which might brake stuff for people. I am not sure that we want that. Can we check if the `api` exposes the config in order to check if Ecip1099 is enabled?

ziogaschr: Maybe using this `api.b.ChainConfig().GetEthashECIP1099Transition()`?

consensus/ethash/ethash.go
```diff
@@ -792,3 +792,13 @@ func (ethash *Ethash) APIs(chain consensus.ChainHeaderReader) []rpc.API {
 func SeedHash(epoch uint64, epochLength uint64) []byte {
 	return seedHash(epoch, epochLength)
 }
+
+// CalcEpochLength returns the epoch length for a given block number (ECIP-1099)
+func CalcEpochLength(block uint64, ecip1099FBlock *uint64) uint64 {
```
ziogaschr, line 797: Do you think `GetEpochLength` fits better in this context?

meowsbits: IMO, it probably does... but `Calc` echos the pre-existing function `calcEpochLength`. Renaming the functions would add diffs for trivial benefits; I'd suggest postponing nice-to-haves like this till later.

meowsbits (approved):

consensus/ethash/sealer.go
meowsbits: Done in subsequent commits.

PR #239, opened 2020-11-16
Update bad cache & dataset hashes

iquidus:
The original hashes produced for the classic mainnet transition were produced before the uint32 overflow fix. Although at the time of generation the bug was unknown, these future datasets are large enough to trigger the overflow so were impacted.

* This PR updates ecip-1099 epoch 195 bad dataset hash.
* Adds bad cache and dataset hashes for ecip-1099 epoch 196.
* Refactors cacheSize & datasetSize, removing the now unused block argument.

### DIFF
--- a/consensus/ethash/algorithm.go
+++ b/consensus/ethash/algorithm.go
@@ -68,7 +68,7 @@ func calcEpoch(block uint64, epochLength uint64) uint64 {
 
 // cacheSize returns the size of the ethash verification cache that belongs to a certain
 // block number.
-func cacheSize(block uint64, epoch uint64) uint64 {
+func cacheSize(epoch uint64) uint64 {
 	if epoch < maxEpoch {
 		return cacheSizes[int(epoch)]
 	}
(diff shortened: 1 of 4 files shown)


ziogaschr:
LGTM. Why are we handling epoch 196 now while we didn't before @iquidus?

meowsbits:
How is this related to #236 ?

iquidus:
> LGTM. Why are we handling epoch 196 now while we didn't before @iquidus?

Just as an extra convenience factor for those that update late (in this case initially after the fork), if they have pre-generated further ahead incorrectly. 

> How is this related to #236 ?

Other than both being 1099 related, there's no relation.

ziogaschr (approved):

meowsbits (approved):

meowsbits:
Just btw, I ran the failing test from https://travis-ci.org/github/etclabscore/core-geth/builds/743952468?utm_source=github_status&utm_medium=notification locally and it passed. Just a flake.

Issue #207, opened 2020-10-12
no mining work available yet


q9f:
#### System information

Geth version: `v1.11.15-stable-51c7e458`
OS & Version: Linux

#### Expected behaviour
Should serve `eth_getWork`

#### Actual behaviour
Does not serve `eth_getWork` api unless you turn on the CPU miner with `miner.start(1)`

#### Steps to reproduce the behaviour
1. launch `geth --http`
2. point `ethminer` to `getwork://127.0.0.1:8545`
3. get `no mining work available yet`

```
WARN [10-12|09:36:48.783] Served eth_getWork                       conn=127.0.0.1:47976 reqid=1 t="57.828µs" err="no mining work available yet"
```

meowsbits:
Running with `geth --mine --http` should fix this (add the `--mine` flag).

meowsbits:
@q9f Does this resolve the issue for you?

q9f:
Yes. My point was that I don't want to mine on CPU but on GPU, so I do not want the `--mine` flag. If this is intended behaviour, then we can close this though.

meowsbits:
Yea, this is intended behavior. Mining with CPU requires `--miner.threads=<1+>`.

q9f:
> Yea, this is intended behavior. Mining with CPU requires `--miner.threads=<1+>`.

Just to be clear, I want to mine on GPU not CPU.

PR #243, opened 2020-11-25
Add an Ethash Fake sealing option for `—dev` mode, using `—dev.ethash`

### LINKED ISSUE #188: feature: add an ethash sealing option for --dev mode
Rel https://github.com/etclabscore/core-geth/issues/187

Currently `--dev` mode uses Clique consensus. This ticket documents a request for an Ethash option for the developer mode.

This would enable
- block production with a locked or absentee etherbase
- Ethash-sealed block consensus field values (eg. difficulty, mix, nonce)

This feature (like `--dev` mode as-is) is essentially an iteration of a convenience/preset flag; the functionality can be created/replicated with existing tools and options by following the pattern of setting up a local or custom config and network.

ziogaschr:
Implement the requested feature at #188.


cmd/utils/flags.go
```diff
@@ -187,7 +187,11 @@ var (
 	}
 	DeveloperPeriodFlag = cli.IntFlag{
 		Name:  "dev.period",
-		Usage: "Block period to use in developer mode (0 = mine only if transaction pending)",
+		Usage: "Block period for proof-of-authority network to use in developer mode (0 = mine only if transaction pending)",
+	}
+	DeveloperEthashFlag = cli.BoolFlag{
+		Name:  "dev.ethash",
```
ziogaschr: @meowsbits do you think `--dev.pow` makes more sense.

Also, I was thinking if anyone might want to test a flow having the ethash and not the fake one. In this case, we can allow this cli param to pass a value like `ethash`, `fakePoW`.

params/genesis.go
```diff
@@ -86,14 +87,19 @@ func DefaultYoloV1GenesisBlock() *genesisT.Genesis {
...
 // DeveloperGenesisBlock returns the 'geth --dev' genesis block. Note, this must
 // be seeded with the
-func DeveloperGenesisBlock(period uint64, faucet common.Address) *genesisT.Genesis {
+func DeveloperGenesisBlock(period uint64, faucet common.Address, useEthash bool) *genesisT.Genesis {
 	// Override the default period to the user requested one
-	config := *AllCliqueProtocolChanges
-	config.Clique.Period = period
+	var config *goethereum.ChainConfig
+	if useEthash {
+		config = AllEthashProtocolChanges
```
ziogaschr: For the `--dev` mode to work automatically from the command line, the user has to pass `--miner.threads > 0`, I was thinking that in the case of developer chain we might want to fix the auto-mining logic by doing this check:

```
if miner.threads == 0
then setThreads = 1
```

cmd/utils/flags.go
meowsbits: Oh yea, interesting line of thought. Following this... What happens with `--miner.threads` values? With my first implementation, this value was co-opted to configure the emission rate.

I'm remembering now that `--fakepow` is an available flag... since that _can_/_could_ be used in conjunction with `--dev.[ethash|pow]`, maybe...

`--dev.[ethash|pow]` enables `--fakepow` by default, since not requiring actual CPU dedication seems usually desireable, making `--miner.threads` configure the emission rate in seconds. This could be overridden with `--fakepow=false`, where `--miner.threads` would revert to configuring the number of dedicated CPUs. 

In general, if we can pick sane defaults that a constrained `--dev.[ethash|pow]` would configure, and then allow adjacent/existing flags to handle further, overriding configuration, then I think we can avoid a rabbit hole of a plethora of development modes (which are really just a bundle of configuration options (flag and genesis)).

consensus/ethash/sealer.go
```diff
@@ -46,18 +47,53 @@ var (
...
 )
 
+// makeFakeDelay uses the ethash.threads value as a mean time (lambda)
+// for a Poisson distribution, returning a random value from
+// that discrete function. I think a Poisson distribution probably
+// fairly accurately models real world block times.
+// Note that this is a hacky way to use ethash.threads since
+// lower values will yield faster blocks, but it saves having
+// to add or modify any more code than necessary.
+func (ethash *Ethash) makeFakeDelay() float64 {
```
meowsbits: @ziogaschr What do you think about breaking this into two PRs:

- an add `fakeDelay` feature which creates a fake-PoW feature that emits blocks following this distribution (no CPUs and configurable, real-world modeling rate)

- add the `dev.[ethash|pow]` flag which would activate this configuration

consensus/ethash/sealer.go
```diff
@@ -46,18 +47,53 @@ var (
...
+	p := distuv.Poisson{
```
meowsbits: Also we may want to rethink the "fake delay" language. IIRC there is a "fake delayer" for testing which basically adds a `time.Sleep` for block verification (but without actually running verification). Disambiguation here may be good.

params/genesis.go
meowsbits: This seems reasonable. The whole idea of the `--dev` mode is provide a practical configuration bundle.

meowsbits (dismissed):
Requesting changes rel https://github.com/etclabscore/core-geth/pull/243/files#r531618009, where I think this should be 2 change sets. Open to be argued otherwise if opinions differ.

consensus/ethash/sealer.go
ziogaschr: Sounds reasonable. Do we want to re-open #180 and check if it's mergable? Then change this one to only include the `--dev.[ethash.pow]` flag?

consensus/ethash/sealer.go
ziogaschr: hmm, this simulates something similar to the difficulty. So will it make sense to name it **similar** to `makeFakeDifficultySimulation`?

cmd/utils/flags.go
ziogaschr: Great ideas. So this will leave us with:

- `--dev.[ethash|pow]`
- `--fakepow`
- `--miner.threads`

Which then can be used like:

- `--dev.[ethash|pow] --miner.threads 1`, this will use ethash
- `--dev.[ethash|pow] --fakepow --miner.threads 1`, this will fakepow (we should document this nicely in the `--dev.[ethash|pow]`cli arg)

where `--miner.threads` if not set will default to:

1. follow the logic used on miner
2. set to **1** if above logic is complicated and coupled to a module

---
What do you think @meowsbits

consensus/ethash/sealer.go
meowsbits: Yea, SGTM. Though I think we'll want to modify 180 a bit to use a dedicated Ethash `Mode` instead of co-opting/re-engineering the existing `ethash.config.PowMode == ModeFake || ethash.config.PowMode == ModeFullFake`.

consensus/ethash/sealer.go
meowsbits: So I think we can use the word "Fake" to mean that `mix` and `nonce` don't need to be legit (are skipped during header verification). The "delay" part is the tricky part.

- https://github.com/etclabscore/core-geth/blob/master/consensus/ethash/ethash.go#L622. Here, "delay" means delay on the verification side. This is only used once, in a test, delaying for one `time.Millisecond`.
- The "delay" 180 proposes is at the _emission_ process, and is intended to model real-world emission pattern (ie Poisson).  Maybe a new `Mode` `Ethash#ModePoissonFake` is in order?

cmd/utils/flags.go
BelfordZ: Have I misunderstood, or is `--dev.[ethash|pow]` supposed to be `--dev.[ethash|poa]`?

consensus/ethash/sealer.go
ziogaschr: I am working on adding a new `Mode: Ethash#ModePoissonFake` on #180. The only question so far, is how will this mode gets enabled @meowsbits?

1. not clean, when `--dev --dev.[ethash|pow] --fakepow` instead of `Ethash#ModeFake` always use `Ethash#ModePoissonFake`. No other way to enable the Poisson mode.
2. introduce a new cli flag `--fakepoissonpow`.
3. change `--fakepow` to string arg accepting following modes [fake|poisson|full].

cmd/utils/flags.go
ziogaschr: By default you will use `--dev` which will enable PoA. If you prefer to have a PoW chain then you will use `--dev --dev.[ethash|pow]`.

Do you have other preferences/recommendations @BelfordZ?

meowsbits: We (I?) were using `[ethash|pow]` to signal that the term was undecided (and that they're basically equivalent).  Leaning to `pow` now for Etchash vs. Ethash considerations.

consensus/ethash/sealer.go
ziogaschr: A `Mode Ethash#ModePoissonFake` has been added at #180.
The enablement of this mode has to be decided, and it will probably happen in this PR ([related comment](https://github.com/etclabscore/core-geth/pull/180#discussion_r533712441)).

params/genesis.go
ziogaschr: Resolved by 28f9f70

consensus/ethash/sealer.go
meowsbits: Yea. Tricky one. We're into what seems like sub-flag territory. Dev mode, with pick-your-poison consensus style; and fakepow mode, with pick-your-poisson emission style. Both have existing defaults. 

Here's what I suggest, for the sake of keeping the CLI API fairly stable, and using consistent namespacing.

- `--dev` behavior is unchanged. Uses Clique.
- `--dev.pow`. A new flag. A new mode. Incompatible with `--dev`.
- `--fakepow`. Behavior is added so that this flag actually does something when `geth` runs. (_ie_ the `SetEthash` addition). Otherwise behavior is unchanged.
- `--fakepow.poisson`. A new flag. A new fake-consensus type. Incompatible with `--fakepow`, but available and useable in the same places and ways.

This leaves the decision What ethash mode to use as a default for `--dev.pow`? I am suggesting to use not-fake PoW as the default, but, as @BelfordZ suggested, toggle the equivalent of `--mine --miner.threads=1`. The `--fakepow[.poisson]` flag(s) would be _applicable_, just not be enabled by default. 

This keeps the mode as "simple" as possible while also keeping options open for the user.

cmd/geth/main.go
```diff
@@ -510,6 +511,9 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend) {
 			threads = ctx.GlobalInt(utils.LegacyMinerThreadsFlag.Name)
 			log.Warn("The flag --minerthreads is deprecated and will be removed in the future, please use --miner.threads")
 		}
+		if threads == 0 && ctx.GlobalBool(utils.DeveloperFlag.Name) {
```
meowsbits: Instead of checking `threads == 0`, maybe should check `!ctx.GlobalIsSet(utils.MinerThreadsFlag.Name) && !ctx.GlobalIsSet(utils.LegacyMinerThreadsFlag.Name)`. Think so?

cmd/utils/flags.go
```diff
@@ -192,7 +192,11 @@ var (
 	}
 	DeveloperPeriodFlag = cli.IntFlag{
 		Name:  "dev.period",
-		Usage: "Block period to use in developer mode (0 = mine only if transaction pending)",
+		Usage: "Block period for proof-of-authority network to use in developer mode (0 = mine only if transaction pending)",
+	}
+	DeveloperPoWFlag = cli.BoolFlag{
+		Name:  "dev.pow",
+		Usage: "Ephemeral proof-of-work network with a pre-funded developer account, mining enabled",
```
meowsbits, line 199: We need an incompatibility check (with `--dev` and `--dev.period`) for this new flag. `utils` has a helper function for that somewhere.

meowsbits: `CheckExclusive`

ziogaschr: I've done this already (locally) :)

params/genesis.go
meowsbits: Is the `AllEthashProtocolChanges` config going to be OK to use here? There are some considerations that Clique doesn't have to make for PoW, eg. difficulty (and difficulty bomb), block rewards.

cmd/geth/main.go
ziogaschr: If you check from line 509 till here it covers flag check, so I thought it will be cleaner this way. What do you think? At the end what we want is to deal with `threads=0`, on the other hand this will not allow user forcing `--miner.threads=0`

params/genesis.go
ziogaschr: Good point. I haven't thought of this. I thought it will work as expected based on [initial PoW->Clique change](https://github.com/ethereum/go-ethereum/pull/15323).

Block rewards work fine. I wonder if we should care about difficulty bomb in `--dev.pow]` mode, or a good use case for it

consensus/ethash/sealer.go
ziogaschr: @meowsbits all the above have been added. The only consideration is that `--dev.pow --fakepow` doesn't sealing as I reverted your changes at #180 to the original code from go-ethereum and added your changes only for the new `ModePoissonFake`. Check [here](https://github.com/etclabscore/core-geth/blob/c3b5cb551ab95de3832a5e5e6808296458fc0fb1/consensus/ethash/sealer.go#L70-L106).

I think I misunderstood your intentions on #180. Do you want me to bring back your change so as sealing works for `--fakepow`?

meowsbits: > `--dev.pow --fakepow` doesn't sealing

Does this mean that it won't automatically start mining blocks? If so, then yes, I'd suggest we change that it does start automatically mining. Or, I'm not sure what you mean by sealing doesn't work for `fakepow`...?

cmd/geth/main.go
meowsbits: I would prefer to allow users the full extent of potential control of the system, even at the cost of a little code cleanliness.  So I think we should let them be able to set `miner.threads=0`. Come to think of it, can't `miner.threads=0` be used to turn off CPU mining while still allowing GPU mining via `eth_getWork`? (See https://github.com/etclabscore/core-geth/issues/207)

params/genesis.go
meowsbits: Dev mode shouldn't ever activate the difficulty bomb, in my opinion. And since core-geth has the ECIP available to permanently disable it, I think it should do so. We'll need to double check the `difficulty` field in the genesis block, too, to make sure it is at the `params.MinimumDifficulty` (or whatever it is (that's pseudo-code ;)). 

The thought I had in mind with the block rewards was in regards to differences available for ETH vs. ETC. For example, `AllEthashProtocolChanges` will set the reward to the ETH _Istanbul_ (permanently), while ETC might, say, set it to Era 1 of [ECIP1017](https://github.com/ethereumclassic/ECIPs/blob/master/_specs/ecip-1017.md), and would then need to establish some arbitrary value for the Era Rounds length (ie probably not 500000)... maybe `math.MaxUint64`.

ziogaschr: Nice one, thanks for analysing that, it really helps me to grasp the whole system. 
I will have a look at it. It will be nice if you can give some tips on it @meowsbits , so as I learn and not spending so much time.

consensus/ethash/sealer.go
ziogaschr: Updating here based on our discussion.

`--fakepow` is used only for the verify process and disables sealing of blocks (actually it create dump blocks), this should not be changed as it will affect the actual use case of `--fakepow`. For this reason `--dev.pow --fakepow` will not be supported as implemented https://github.com/etclabscore/core-geth/pull/243/commits/6eb9ce9e7b0ef097880deb3b65b616959a549077.

The supported modes will be:

1. `--dev.pow`: starting an Ethash Ephemeral chain
2. `--dev.pow --fakepow.poisson`: starting an Ethash Ephemeral chain, without verification of blocks and with mining delay (Poisson) based on --miner.threads

params/genesis.go
meowsbits: Sure, check out https://github.com/etclabscore/core-geth/pull/243/commits/f4530b2c1cab406bc45f35f4429d1573490d1be5

meowsbits (approved):

params/genesis.go
ziogaschr: Thanks, very informative. EIPs used for block rewards and difficulty bomb noted. Own you a beer

PR #244, opened 2020-11-26
Allow Ethereum protocol version to be set user configurable

### LINKED ISSUE #109: create an --eth-protocol=[65|64|63] flag?
Just an idea to allow the Ethereum Protocol version to be user configurable, which would provide a simple fix for issues like the ones Besu and ~~Parity~~ OpenEthereum are having on ETC right now. https://github.com/openethereum/openethereum/issues/11744

https://github.com/etclabscore/core-geth/blob/master/eth/protocol.go#L43

ziogaschr:
Adds a new cli arg to `geth` for configuring the supported Ethereum protocol versions.
The new cli arg is `--ethprotocols=[value]` which accepts comma separated string of the supported protocol versions **[65|64|63]**.

This PR resolves #109.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -27,6 +27,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"text/tabwriter"
@@ -140,6 +141,11 @@ var (
 		Name:  "networkid",
 		Usage: "Explicitly set network id (integer)(For testnets: use --ropsten, --rinkeby, --goerli, --kotti, --mordor, --yolov1 instead)",
 		Value: eth.DefaultConfig.NetworkId,
+	}
+	EthProtocolsFlag = cli.StringFlag{
+		Name:  "ethprotocols",
+		Usage: "Sets the Ethereum Protocol versions (65|64|63) (default=65,64,63)",
+		Value: "",
 	}
 	ClassicFlag = cli.BoolFlag{
 		Name:  "classic",
@@ -1724,6 +1730,40 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 		cfg.NetworkId = ctx.GlobalUint64(NetworkIdFlag.Name)
 	} else if cfg.Genesis != nil {
 		cfg.NetworkId = *cfg.Genesis.GetNetworkID()
+	}
+
+	if ctx.GlobalIsSet(EthProtocolsFlag.Name) {
+		protocolVersions := SplitAndTrim(ctx.GlobalString(EthProtocolsFlag.Name))
+		if len(protocolVersions) > 0 {
+			for _, versionString := range protocolVersions {
+				version, err := strconv.ParseUint(versionString, 10, 0)
+				if err != nil {
+					continue
+				}
+
+				isValid := false
+				for _, proto := range eth.DefaultProtocolVersions {
+					if proto == uint(version) {
+						isValid = true
+					}
+				}
+
+				if !isValid {
+					Fatalf("--%s must be one of %s", EthProtocolsFlag.Name, strings.Join(strings.Fields(fmt.Sprint(eth.DefaultProtocolVersions)), ","))
+				}
+				cfg.ProtocolVersions = append(cfg.ProtocolVersions, uint(version))
+			}
+		}
+
+		// sort protocol version desceding
+		sort.Slice(cfg.ProtocolVersions, func(i, j int) bool { return cfg.ProtocolVersions[i] > cfg.ProtocolVersions[j] })
+		fmt.Printf("ProtocolVersions: %#v\n", cfg.ProtocolVersions)
+	}
+
+	// set default protocol versions
+	if len(cfg.ProtocolVersions) == 0 {
+		fmt.Printf("default ProtocolVersions: %#v\n", cfg.ProtocolVersions)
+		cfg.ProtocolVersions = eth.DefaultProtocolVersions
 	}
 
 	// Set DNS discovery defaults for hard coded networks with DNS defaults.
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -143,7 +143,7 @@ var (
 		Value: eth.DefaultConfig.NetworkId,
 	}
 	EthProtocolsFlag = cli.StringFlag{
-		Name:  "ethprotocols",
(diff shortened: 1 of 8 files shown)


cmd/utils/flags.go
```diff
@@ -1726,6 +1732,38 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
...
+						isValid = true
+					}
+				}
+
+				if !isValid {
+					Fatalf("--%s must be one of %s", EthProtocolsFlag.Name, strings.Join(strings.Fields(fmt.Sprint(eth.DefaultProtocolVersions)), ","))
+				}
+				cfg.ProtocolVersions = append(cfg.ProtocolVersions, uint(version))
+			}
+		}
```
meowsbits: (else if `len(protocolVersions) == 0`) should `Fatalf` in same way as `!isValid`? Force user to provide valid values if the flag is set.

cmd/utils/flags.go
```diff
@@ -1726,6 +1732,38 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
...
+
+		// sort protocol version desceding
+		sort.Slice(cfg.ProtocolVersions, func(i, j int) bool { return cfg.ProtocolVersions[i] > cfg.ProtocolVersions[j] })
```
meowsbits: > // Protocol versions are the supported versions of the eth protocol (first is primary)
> https://github.com/etclabscore/core-geth/pull/244/files#diff-cb10e99654aa32525bbb358ea722f79e26da92b3ecdc70e29ce1d711f1be3de4R116

Do we really want to sort these?

cmd/utils/flags.go
```diff
@@ -141,6 +142,11 @@ var (
 		Usage: "Explicitly set network id (integer)(For testnets: use --ropsten, --rinkeby, --goerli, --kotti, --mordor, --yolov1 instead)",
 		Value: eth.DefaultConfig.NetworkId,
 	}
+	EthProtocolsFlag = cli.StringFlag{
+		Name:  "ethprotocols",
+		Usage: "Sets the Ethereum Protocol versions (65|64|63) (default=65,64,63)",
```
meowsbits: Since "first is primary", maybe should add a note here mentioning that?

cmd/utils/flags.go
```diff
@@ -141,6 +142,11 @@ var (
 		Usage: "Explicitly set network id (integer)(For testnets: use --ropsten, --rinkeby, --goerli, --kotti, --mordor, --yolov1 instead)",
 		Value: eth.DefaultConfig.NetworkId,
 	}
+	EthProtocolsFlag = cli.StringFlag{
+		Name:  "ethprotocols",
```
meowsbits: What about `eth.protocols`? It seems like a general pattern is `<service>.<value>`.. ?

cmd/utils/flags.go
ziogaschr: I found a place in the network layer where on the server side logic they peek the highest one. But yeah, let's better not break the primary logic here by just a consumption. Fixing

meowsbits: Interesting. Is "a place in the network layer where on the server side logic they peek the highest one" worth spinning off into own investigate/fix issue?

ziogaschr: No, as the logic there was in a different mindset. Will try to find this again and let you know in chat. Let's merge this one once ready

cmd/utils/flags.go
```diff
@@ -1726,6 +1731,37 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 		cfg.NetworkId = *cfg.Genesis.GetNetworkID()
 	}
 
+	if ctx.GlobalIsSet(EthProtocolsFlag.Name) {
+		protocolVersions := SplitAndTrim(ctx.GlobalString(EthProtocolsFlag.Name))
+		if len(protocolVersions) > 0 {
+			for _, versionString := range protocolVersions {
+				version, err := strconv.ParseUint(versionString, 10, 0)
+				if err != nil {
+					continue
```
meowsbits: Let's handle this error, forcing the user to provide valid values (and loudly erroring if not).

cmd/utils/flags.go
```diff
@@ -1726,6 +1731,37 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
...
+				}
+
+				isValid := false
+				for _, proto := range eth.DefaultProtocolVersions {
+					if proto == uint(version) {
+						isValid = true
```
meowsbits: Can `break` here, or?

cmd/utils/flags.go
meowsbits: Or ignore the error altogether, and assume that `version (==0)` won't be an existing (and thus won't be valid) as a `DefaultProtocolVersion`.

meowsbits: But I think returning the error lets us provide the user with more useful context.

cmd/utils/flags.go
```diff
@@ -1726,6 +1731,37 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
...
+					}
+				}
+
+				if !isValid {
+					Fatalf("--%s must be comma separated list of %s", EthProtocolsFlag.Name, strings.Join(strings.Fields(fmt.Sprint(eth.DefaultProtocolVersions)), ","))
+				}
+				cfg.ProtocolVersions = append(cfg.ProtocolVersions, uint(version))
```
meowsbits: Do we need to do any further sanitation here? For example, should `--eth.protocols=65,65,65` work?

cmd/utils/flags.go
```diff
@@ -1726,6 +1731,37 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 		cfg.NetworkId = *cfg.Genesis.GetNetworkID()
 	}
 
+	if ctx.GlobalIsSet(EthProtocolsFlag.Name) {
+		protocolVersions := SplitAndTrim(ctx.GlobalString(EthProtocolsFlag.Name))
+		if len(protocolVersions) > 0 {
```
meowsbits: We can flip this logic to remove a thick condition (and be a little more Go-idiomatic):

```
if len(protocolVersions) == 0
  // handle the Fatalf for bad usage

for _, versionString := range protocolVersions
```

meowsbits (approved):
LGTM

PR #246, opened 2020-12-04
Convert YoloV2 flags to EIP-2315 and EIP-2929

ziogaschr:


core/vm/runtime/runtime.go
```diff
@@ -113,17 +113,14 @@ func Execute(code, input []byte, cfg *Config) ([]byte, *state.StateDB, error) {
...
-			cfg.State.AddAddressToAccessList(address)
-			for _, addr := range vmenv.ActivePrecompiles() {
-				cfg.State.AddAddressToAccessList(addr)
-				cfg.State.AddAddressToAccessList(addr)
-			}
+	if cfg.ChainConfig.IsEnabled(cfg.ChainConfig.GetEIP2929Transition, vmenv.BlockNumber) {
+		cfg.State.AddAddressToAccessList(cfg.Origin)
+		cfg.State.AddAddressToAccessList(address)
+		for addr := range vm.PrecompiledContractsForConfig(cfg.ChainConfig, vmenv.BlockNumber) {
+			cfg.State.AddAddressToAccessList(addr)
```
ziogaschr, line 120: The duplicated line here, is already being tracked at [go-ethereum](https://github.com/ethereum/go-ethereum/pull/21956)

params/config.go
```diff
@@ -30,8 +30,7 @@ var (
 	RopstenGenesisHash = common.HexToHash("0x41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2d")
 	RinkebyGenesisHash = common.HexToHash("0x6341fd3daf94b748c72ced5a5b26028f2474f5f00d824504e4fa37a75767e177")
 	GoerliGenesisHash  = common.HexToHash("0xbf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a")
-	// TODO: update with yolov2 values
-	YoloV2GenesisHash = common.HexToHash("0x498a7239036dd2cd09e2bb8a80922b78632017958c332b42044c250d603a8a3e")
+	YoloV2GenesisHash  = common.HexToHash("0xc3fd235071f24f93865b0850bd2a2119b30f7224d18a0e34c7bbf549ad7e3d36") // TODO: check if yolov2 is correct
```
ziogaschr: At go-ethereum they left this on yolov1 until they get the new value. Though we have a testcase that fails. So I temporarily used the output of the test to make it happy

params/types/coregeth/chain_config.go
```diff
@@ -163,7 +163,7 @@ type CoreGethChainConfig struct {
 	EIP1706FBlock *big.Int `json:"eip1706FBlock,omitempty"`
 
 	// https://github.com/ethereum/EIPs/pull/2537: BLS12-381 curve operations
-	EIP2537FBlock *big.Int `json:"eip2315FBlock,omitempty"`
+	EIP2537FBlock *big.Int `json:"eip2537FBlock,omitempty"`
```
ziogaschr, line 166: This seems to be a mistype which I fixed. Correct me if it was intentionally

params/types/parity/parity.go
```diff
@@ -106,8 +106,10 @@ type ParityChainSpec struct {
 		EIP1344Transition         *ParityU64 `json:"eip1344Transition,omitempty"`
 		EIP1884Transition         *ParityU64 `json:"eip1884Transition,omitempty"`
 		EIP2028Transition         *ParityU64 `json:"eip2028Transition,omitempty"`
+		EIP2315Transition         *ParityU64 `json:"eip2315Transition,omitempty"`
 		EIP2537Transition         *ParityU64 `json:"eip2537Transition,omitempty"`
 		EIP1706Transition         *ParityU64 `json:"-"` // FIXME, when and if i'm implemented in Parity
+		EIP2929Transition         *ParityU64 `json:"-"` // FIXME, when and if i'm implemented in Parity
```
ziogaschr, line 112: We have a test case on forks (TestEquivalent_Features) that fails on this transition. The thing is that OpenEthereum is not going to implement this EIP as they are out of resources at this moment. 

What is the best way to handle this test case @meowsbits?

params/config.go
holiman: Not really, we just chose to use the exact same genesis (alloc + time but not chainId) for yolov2 as yolov1.

ziogaschr: Thanks @holiman for jumping in, this clears up everything. I will fix anything needed to make the test pass then.

cmd/puppeth/wizard_genesis.go
```diff
@@ -240,8 +240,8 @@ func (w *wizard) manageGenesis() {
 		w.conf.Genesis.Config.SetEIP145Transition(w.readDefaultUint64P(*w.conf.Genesis.Config.GetEIP145Transition()))
 
 		fmt.Println()
-		fmt.Printf("Which block should YOLOv1 (EIP2537) come into effect? (default = %v)\n", w.conf.Genesis.Config.GetEIP2537Transition())
-		w.conf.Genesis.Config.SetEIP2537Transition(w.readDefaultUint64P(*w.conf.Genesis.Config.GetEIP2537Transition()))
+		fmt.Printf("Which block should YOLOv2 come into effect? (default = %v)\n", w.conf.Genesis.Config.GetEIP2929Transition())
```
ziogaschr, line 243: @meowsbits there is not really an EIP for YoloV2 yet, so I used the EIP2929 as the flag for that

params/types/coregeth/chain_config.go
meowsbits: You are correct, it was my mistake

eth/api_tracer.go
```diff
@@ -597,22 +598,17 @@ func (api *PrivateDebugAPI) standardTraceBlockToFile(ctx context.Context, block
...
-			if yolov2 := config.Overrides.YoloV2Block; yolov2 != nil {
-				chainConfig.YoloV2Block = yolov2
-				canon = false
+		// Copy the config, to not screw up the main config
+		// Note: the Clique-part is _not_ deep copied
+		chainConfigCopy := new(ctypes.ChainConfigurator)
+		*chainConfigCopy = chainConfig
+		chainConfig = *chainConfigCopy
+		if eip2929 := config.Overrides.GetEIP2929Transition(); eip2929 != nil {
+			if err := chainConfig.SetEIP2929Transition(eip2929); err != nil {
```
meowsbits, line 607: So we have to use `config.Overrides.GetEIP2929Transition(); eip2929 != nil` to see if YoloV2 is activated (using an indicator EIP as a proxy for all associated EIPs for that fork.

But don't we have to do activations like `chainConfig.SetEIP2929Transition(eip2929)` for _all_ the EIPs in that fork (ie `chainConfig.SetEIP2315Transition(eip2929)`, etc.)?

ziogaschr: Good one. Yes we have to handle all Yolo EIPs this way individually, I am not sure if we should use an indicator EIP to read the fork block number and use it to set all the EIPs on this block. The reason is that on some chain we might want to include only some of the EIPs and not all.

I will add `EIP2315` and `EIP2537` for now.

ziogaschr: The overrides were actually used for producing an analysis for EIP-2929 and not being used in tests or anywhere else in the code for now. You can check [here](https://github.com/ethereum/go-ethereum/pull/21509#issue-477308710)

ziogaschr: Resolved with https://github.com/etclabscore/core-geth/pull/246/commits/75af684928f0041d9c5df88776f3e956a508eda4

ziogaschr: I am re-thinking of it @meowsbits .
1. What they wanted to test was actually only the EIP2929 implications, so the rest can be removed
2. We can even remove this block from here and consider it tested. Later on we can bring it back if we need to test similar E[C]IPs

params/types/parity/parity.go
meowsbits: I'm guessing `func (spec *ParityChainSpec) SetEIP2929Transition(n *uint64) error {` needs to actually set `EIP2929Transition`, in the same way as EIP2315 is handled for Parity.

params/types/parity/parity_configurator.go
```diff
@@ -985,6 +985,26 @@ func (spec *ParityChainSpec) SetEthashECIP1099Transition(n *uint64) error {
...
+
+func (spec *ParityChainSpec) GetEIP2929Transition() *uint64 {
+	return nil
+}
+
+func (spec *ParityChainSpec) SetEIP2929Transition(n *uint64) error {
+	if n == nil {
+		return nil
+	}
+	return ctypes.ErrUnsupportedConfigFatal
```
meowsbits, line 1005: Rel https://github.com/etclabscore/core-geth/pull/246/files#r536073142

eth/api_tracer.go
meowsbits: SGTM. Nice investigation!

ziogaschr: @meowsbits so **leave the code as is** or go with **1** or **2**. I prefer 1.

params/types/parity/parity.go
meowsbits: Resolved with https://github.com/etclabscore/core-geth/pull/246/commits/f3aff0ec9c2ec7342973e19f739c881c226bd867

eth/api_tracer.go
meowsbits: Yes, 1 SGTM.  In general I want to own and persist a superset of the tests from ethereum/go-ethereum (ie. never _only_ rely on them to have tested something, we need to own all the tests... and more ;).

params/types/parity/parity.go
ziogaschr: This resolves our test but I though of leaving it unsupported based on the comment [here](https://docs.google.com/spreadsheets/d/1BomvS0hjc88eTfx1b8Ufa6KYS3vMEb2c8TQ5HJWx2lc/edit#gid=408811124) saying that they are not going to implement this.

cmd/puppeth/wizard_genesis.go
meowsbits: Yeap, here's the EIP-as-fork-indicator scenario again. This might eventually warrant it's own logic somewhere (a fork :: indicator EIP function). Another important place where this happens is here, for example: https://github.com/etclabscore/core-geth/blob/master/core/vm/evmc.go#L302

ziogaschr: Thanks for sharing the evmc link. Good to know that something like this exists.

eth/api_tracer.go
ziogaschr: Resolves with https://github.com/etclabscore/core-geth/pull/246/commits/26b814975e1c7cdf3afb9f81756c1c0f60446f21.
Which reverts the previous resolving commit :)

params/types/parity/parity.go
meowsbits: Ah, roger that. I've reverted that commit and just modified the tests to skip Parity :: [Berlin|YOLO] tests. See https://github.com/etclabscore/core-geth/pull/246/commits/186e3cb59becdd812fa4c57da6ec3e4458e899dc and subsequent.

meowsbits (approved):

Issue #240, opened 2020-11-16
faucet (lightclient) @Mordor: bootnode, peering issues


meowsbits:
#### System information

```
./build/bin/geth version
CoreGeth
Version: 1.11.19-unstable
Git Commit: 061c097c8420a8397a0a31327f11bf13108978ba
Git Commit Date: 20201113
Architecture: amd64
Protocol Versions: [65 64 63]
Go Version: go1.13.5
Operating System: linux
GOPATH=/root/go
GOROOT=go
```

#### Expected behaviour

Finds peers by default.

#### Actual behaviour

Not peering.

#### Steps to reproduce the behaviour

```
[Unit]
Description=FaucetMordor

[Service]
Type=simple
User=root
Restart=always
RestartSec=3
TimeoutStopSec=10min
ExecStart=/bin/sh -c '/.../build/bin/faucet -apiport 80 -loglevel=6 -chain.mordor -network=7 -faucet.name="Mordor" -account.json=/path/to/account.json -account.pass=/path/to/pass.txt -faucet.amount=10 -ethport 30305'

[Install]
WantedBy=default.target


```

#### Backtrace

Note that the `bootnodes=` value is empty. This means that bootnodes by enode should be replaced by DNS, or a value should be provided.

Also note that Mordor has a scarcity of light peers available (maybe only one?), so that may be relevant here as well. 

````
Nov 16 18:32:08 sf-etclabs systemd[1]: Started FaucetMordor.                                                                                                                                                                                                                    
Nov 16 18:32:09 sf-etclabs sh[27642]: INFO [11-16|18:32:09.071] configured chain/net config              network id=7 bootnodes= chain config="NetworkID: 7, ChainID: 63 Engine: ethash ECBP1100: 2380000 EIP1014: 301243 EIP1052: 301243 EIP1108: 999983 EIP1344: 999983 EIP140
: 0 EIP145: 301243 EIP150: 0 EIP152: 999983 EIP155: 0 EIP160: 0 EIP161abc: 0 EIP161d: 0 EIP170: 0 EIP1884: 999983 EIP198: 0 EIP2028: 999983 EIP211: 0 EIP212: 0 EIP213: 0 EIP214: 0 EIP2200: 999983 EIP2: 0 EIP658: 0 EIP7: 0 EthashECIP1017: 0 EthashECIP1041: 0 EthashECIP1099
: 2520000 EthashEIP100B: 0 EthashHomestead: 0 "                                                                                                                                                                                                                                 
[...]

PR #249, opened 2020-12-08
Fix/faucet discovery dns defaults

meowsbits:
No DNS discovery defaults were getting set
because you can't test on pointer equivalence
, which was being done on the genesis blocks.

Instead, compare on Hash equivalence
(and handle disambiguation for ETC vs. ETH
 using chain ID (61 vs. 1)).

Adds a couple log lines at start up for
more helpful debug logging.

Date: 2020-12-08 06:53:29-06:00


meowsbits:
Rel https://github.com/etclabscore/core-geth/issues/240

meowsbits:
Fixes #251 
Fixes #240

ziogaschr (approved):
LGTM

PR #257, opened 2020-12-18
fix: update classic bootnodes

stevanlohja:
Removed Parity boot node and added ETC Coop boot-nodes (which are Besu nodes).

We recently updated Hyperledger Besu boot-nodes to https://github.com/hyperledger/besu/pull/1728


meowsbits (changes requested):
Thanks! Great to have some Besu nodes spoken for. Please fix the [formatting issue](https://travis-ci.org/github/etclabscore/core-geth/jobs/750456277#L250) and this'll be gg.

meowsbits (approved):

Issue #258, opened 2020-12-21
feature: allow faucet to attach to client


meowsbits:
#### Expected behaviour

`faucet` can `attach` to a running instance

#### Actual behaviour

`faucet` can only start and use it's own LES client.

Rel #251

meowsbits:
Done with #269

PR #259, opened 2020-12-21
[WIP] main,ethclient,ethstats: init faucet attach feature

meowsbits:
Rel https://github.com/etclabscore/core-geth/issues/258

This begins implementation of the ref'd feature proposal.

Currently 'ethstats' support is removed because
that package relies heavily on access to a *node.Node.
Eventually I want to move that packages demands to
something that can be satisfied by an API instead.

Date: 2020-12-21 12:16:33-06:00


meowsbits:
Closing deferring to #269

PR #261, opened 2020-12-21
Faucet: Add log when importing faucet signer account

ziogaschr:
This PR adds just a log when the signer account is being imported, as some previous log messages (attached image) might be misleading, and the user might feel that account import failed.

image


meowsbits (approved):
IMO it might be easier to read as `} else if err == nil {` which is the same logical case.

ziogaschr:
> IMO it might be easier to read as `} else if err == nil {` which is the same logical case.

thanks man. Pity one for my side

PR #268, opened 2020-12-28
Feat/fast faucet

meowsbits:
- Creates `-syncmode=[light|full|fast]` flag, with default `light` to preserve default behavior.
- Creates `-datadir` flag to allow configurable datadir placement.


meowsbits:
Note that the last 3 commits (https://github.com/etclabscore/core-geth/pull/268/commits/01bc50d9e82a09588a95577df112528369c0e3d9 ...) merge `master` which includes conflicting changes from #269. These conflicts were committed along with their subsequent resolutions. As a result, this change set should be squash-merged.

cmd/faucet/faucet.go
```diff
@@ -440,6 +446,13 @@ func (f *faucet) startStack(genesis *genesisT.Genesis, port int, enodes []*discv
 	// Assemble the Ethereum light client protocol
 	cfg := eth.DefaultConfig
 	cfg.SyncMode = downloader.LightSync
+	if *syncmodeFlag == "fast" {
+		cfg.SyncMode = downloader.FastSync
+		cfg.ProtocolVersions = eth.DefaultProtocolVersions
+	} else if *syncmodeFlag == "full" {
+		cfg.SyncMode = downloader.FullSync
+		cfg.ProtocolVersions = eth.DefaultProtocolVersions
+	}
```
ziogaschr, line 475: We might want to be explicit here that we don't handle anything else than "light".

Maybe we can add something like:
```go
} else if *syncmodeFlag != "light" {
  return fmt.Error("Supported Sync modes are one of [light | fast | full].")
}
```

ziogaschr (commented):
Please also have a look at https://github.com/etclabscore/core-geth/pull/269/files#r551442004

cmd/faucet/faucet.go
meowsbits: Yea, good call. See https://github.com/etclabscore/core-geth/pull/268/commits/c883a4a221a589ea62f3d2d14669e7435885a5d3, where I've enforced this in `auditFlagUse` and refactored a little to both improve legibility and remove useless `backend` construction (which is only used for `ethstats` reporting).

ziogaschr: Nice refactor man. It's more readable now

ziogaschr (approved):
LGTM

PR #269, opened 2020-12-28
Feat/faucet attach 2

meowsbits:
Supersedes #259. 

- Creates `-attach` flag allowing the faucet program to _not_ run it's own client, instead attaching to an existing one via JSONRPC. Target client API needs to have `eth` and `net` API modules exposed.

- (https://github.com/etclabscore/core-geth/pull/269/commits/c99ed4582f30ed2ca289d43419ffd09a2639be40) _Disables_ `NoDiscovery` setting for managed client P2P configuration. Allowing discovery allows the faucet to not rely exclusively on manually configured bootnodes. Left commented because this differs from upstream... and I'm not sure at this point what the rationale was for disabling p2p discovery in the first place.


cmd/faucet/faucet.go
```diff
@@ -286,7 +306,26 @@ func main() {
...
 
-	ks := keystore.NewKeyStore(filepath.Join(faucetDirFromConfig(genesis.Config), "keys"), keystore.StandardScryptN, keystore.StandardScryptP)
+	// Get the chain id from the genesis or the designated API.
+	// We'll use this to infer the keystore and chain data directories.
+	// NOTE: Relying on having chain id immediately available may be fragile.
+	// IRC eth_chainId can possibly not return the configured value until the block activating the chain id EIP155 is reached.
+	// See the difference in implementation between ethapi/api.go and eth/api.go #ChainID() methods.
+	// There's an issue open about this somewhere at ethereum/xxx.
+	// This could be resolved by creating a -chainid flag to use as a fallback.
+	chainID := uint64(0)
```
meowsbits, line 321: :eyes: (see comment)

ziogaschr: `-chainid` sounds good to have as a fallback.

Maybe we can request the genesis hash from the attached node and compare it locally? (till EIP155)

meowsbits: Yeap, that'll work. I just wrote the code for it (see https://github.com/etclabscore/core-geth/pull/269/commits/b78074862257c7943e18df3912893a01d7ba4a6e and +1), but am now getting the feeling of fix-it-with-a-flag/feature-creep. The only thing this is used for is configuring the data directory namespace. #268 includes a change adding a `-datadir` flag that would provide another way around this potential ambiguity for a user. Otherwise, this ambiguity would only be an issue for users attempting to attach the `faucet` to a (currently) very-unsynced client. I'm starting to feel inclined to just `if chainID == 0 && { log.Crit... }`.

ziogaschr: I like the way https://github.com/etclabscore/core-geth/commit/b78074862257c7943e18df3912893a01d7ba4a6e + 1 handles it. Bravo!!!

cmd/faucet/faucet.go
```diff
@@ -629,7 +725,15 @@ func (f *faucet) apiHandler(w http.ResponseWriter, r *http.Request) {
 			amount = new(big.Int).Div(amount, new(big.Int).Exp(big.NewInt(2), big.NewInt(int64(msg.Tier)), nil))
 
 			tx := types.NewTransaction(f.nonce+uint64(len(f.reqs)), address, amount, 21000, f.price, nil)
-			signed, err := f.keystore.SignTx(f.account, tx, f.config.GetChainID())
+
+			// FIXME(meowsbits): Getting the chain id more than once is redundant and can be optimized.
```
iquidus, line 729: is there a solution for this in mind?

iquidus (approved):

cmd/faucet/faucet.go
meowsbits: Probably there are at least a few ways of handling it. One that comes first to mind is to add a field on `faucet`, eg. `faucet.chainID` since we know that once it's set to a nonzero value it won't change, and the `faucet` is only designed to handle a single network. But IMO this is a nice-to-have optimization since the request will only be made for each faucet transaction... which isn't more than can be expected for a "standard" transaction-making program (at least it's not getting called at every block or anything like that). Because of this I'm OK pushing to a later optimization if need/desire calls for it.

cmd/faucet/faucet.go
```diff
@@ -176,107 +273,33 @@ func main() {
...
-	log.Info("configured chain/net config", "network id", *netFlag, "bootnodes", *bootFlag, "chain config", fmt.Sprintf("%v", genesis.Config))
-
-	// Convert the bootnodes to internal enode representations
-	for _, boot := range utils.SplitAndTrim(*bootFlag) {
-		if url, err := discv5.ParseNode(boot); err == nil {
-			enodes = append(enodes, url)
-		} else {
-			log.Error("Failed to parse bootnode URL", "url", boot, "err", err)
+	} else {
+		log.Info("Attaching faucet to running client")
```
ziogaschr, line 299: @meowsbits this message is a bit misleading. While testing #268 without using neither `-attach` nor a chain `-chain.*` flag. It get's in here saying it failed to attach. While the truth is that it failed to peek a chain.

meowsbits: Nice catch. Please see https://github.com/etclabscore/core-geth/pull/275.

PR #275, opened 2021-01-04
main: fix spurious error log when running without -chain or -attach

meowsbits:
This enforces what was an implicit requirement
that the faucet be configured explicitly for
a chain configuration.

Fixing https://github.com/etclabscore/core-geth/pull/269/files#r551442004,
the bespoke log happens whent neither -attach
nor -chain.xxx is set, the behavior for which should
be undefined.

Date: 2021-01-04 12:37:10-06:00


ziogaschr:
LGTM

PR #293, opened 2021-01-12
eth/filters,ethclient,node: install newSideHeads subscription

### LINKED ISSUE #287: rpc: add newSideBlocks subscription
The RPC API currently supports a few _subscription_ types, eg:
- `newHeads`
- `log`
- `newPendingTransactions`
- `syncing`

This issue proposes adding an `events` method and subscription endpoint for `newSideBlocks` akin to `newHeads`, but where the reported blocks would be those considered _non-canonical_. 

Rel #284

meowsbits:
The newSideHeads subscription work very similarly to
the newHeads subscription; instead, non-canonical blocks are channeled.

Resolves #287

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -236,7 +236,6 @@ func (api *PublicFilterAPI) NewSideBlockFilter() rpc.ID {
 	return headerSub.ID
 }
 
-
 // NewHeads send a notification each time a new (header) block is appended to the chain.
 func (api *PublicFilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, error) {
 	notifier, supported := rpc.NotifierFromContext(ctx)
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -203,6 +203,40 @@ func (api *PublicFilterAPI) NewBlockFilter() rpc.ID {
 	return headerSub.ID
 }
 
+// NewSideBlockFilter creates a filter that fetches blocks that are imported into the chain with a non-canonical status.
+// It is part of the filter package since polling goes with eth_getFilterChanges.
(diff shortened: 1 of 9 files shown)


eth/filters/filter_system.go
```diff
@@ -399,6 +440,10 @@ func (es *EventSystem) lightFilterNewHead(newHeader *types.Header, callBack func
 	}
 }
 
+func (es *EventSystem) lightFilterNewSideHead(header *types.Header, callBack func(*types.Header, bool)) {
+	callBack(header, true)
```
meowsbits, line 444: This is noteworthy; logs from side chain blocks will always be marked `Removed: true`. `newHeads` handling of this uses `Removed` to indicate the same characteristic, but in that context is expected to _have been_ canonical, and is now _removed_ to the side. Although the reuse of this field (and vocabulary) is a little awkward, I think it's the best we can do without modifying the existing API data type.

eth/filters/api.go
```diff
@@ -233,7 +266,37 @@ func (api *PublicFilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, er
 	return rpcSub, nil
 }
 
-// Logs creates a subscription that fires for all new log that match the given filter criteria.
+// NewHeads send a notification each time a new (header) block is appended to the chain.
```
ziogaschr: This comment is wrong

ziogaschr (approved):
LGTM!

meowsbits:
Merged with https://github.com/etclabscore/core-geth/commit/9422e73dcda8aeace7e0c21516a139db63b86a55 on the CLI: 
```
> git merge --squash --commit --e --signoff -S feat/eth-subscribe-newsideblocks-pr
```

PR #294, opened 2021-01-12
build,Makefile: add 'make mkdocs-serve'

meowsbits:
This adds a script and associated Makefile command
to handle installing requirements and starting the
development server for mkdocs.

### DIFF
--- a/build/mkdocs-serve.sh
+++ b/build/mkdocs-serve.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+# This script installs requirements for and starts the live-reloading development server for MkDocs.
+# MkDocs generates a static site from the markdown files in docs/.
+
+py_command=python
+pip_command=pip
+
+setup_py_command() {
+	if [ -z $(which ${py_command}) ]; then
+		py_command=python3
+		pip_command=pip3
+		return
+	fi
+	py_command_major_version=$(${py_command} 2>&1 --version | cut -d' ' -f2 | head -c1)
+	if [ $py_command_major_version != 3 ]; then
+		py_command=python3
+		pip_command=pip3
+	fi
+}
+setup_py_command
+
+set -ex
+${py_command} --version
+${pip_command} --version
+set +x
+
+${pip_command} install -r requirements-mkdocs.txt
+
+mkdocs serve
--- a/build/mkdocs-serve.sh
+++ b/build/mkdocs-serve.sh
@@ -4,18 +4,16 @@
 # MkDocs generates a static site from the markdown files in docs/.
 
 py_command=python
-pip_command=pip
+pip_command="${py_command} -m pip"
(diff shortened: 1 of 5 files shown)


build/mkdocs-serve.sh
```diff
@@ -0,0 +1,30 @@
...
+		pip_command=pip3
+		return
+	fi
+	py_command_major_version=$(${py_command} 2>&1 --version | cut -d' ' -f2 | head -c1)
+	if [ $py_command_major_version != 3 ]; then
+		py_command=python3
+		pip_command=pip3
+	fi
+}
+setup_py_command
```
meowsbits, line 18: This is kind of awkward; assumes that if `python` is v2 then `pip` will be too; assumes that if so, then `pip3` should be available. If `pip` or `pip3` is not available in the `PATH` it will fail with a `command not found` error at https://github.com/etclabscore/core-geth/pull/294/files#diff-14af3ec200ad61f89a2faf7f748f9fc872bdb23f3836662f2b4a21e0108c6032R24.

There is probably a cleaner way of doing this (maybe even a conventional one?).

ziogaschr: You can call any python script using `python -m`, so for example `python -m pip --version`

ziogaschr: Resolved by https://github.com/etclabscore/core-geth/pull/294/commits/946fa8610b26cc4415b1cb632274c825c5e503b3

ziogaschr:
I wonder if we should add the creation of a virtualenv within the script. As it currently installs the packages globally in the system.

Though not all the developers know/like how to use virtualenvs and at the end a `~/.virtualenv/` dir will be created.

meowsbits:
Maybe we should just use `mkdocs` if it's available in the PATH, or `exit 1` with some help message about needing to install `mkdocs` and where to look to do that.

Then we get to have no opinions about Python or environments...

ziogaschr:
No strong preference on this one. I liked the idea of your script as core-geth is a go project and their users might not know much about python, while most systems have python installed.

meowsbits:
I'm OK leaving out`virtualenv` business. The main reason I had _any_ python management stuff in there was to give context (or an excuse) for putting the `python install -r requirements-mkdocs.txt` in the code somewhere for sake of documentation. ~~Will push a change to simplify this (and trim responsibility).~~

_EDIT_: @ziogaschr Just saw your commit https://github.com/etclabscore/core-geth/pull/294/commits/946fa8610b26cc4415b1cb632274c825c5e503b3 -- LGTM.

ziogaschr:
Would it make sense adding this in both README and docs? Something across the lines of using MkDocs and here is how to install it and run it locally.

ziogaschr (approved):
LGTM

Issue #15, opened 2020-02-19
Feat: dependabot-style PRs when the ethereum-json-rpc-spec changes


meowsbits:
Migrated from: https://github.com/etclabscore/multi-geth-fork/issues/103
Original author: @BelfordZ

---

rel #99

ziogaschr:
How do we check for spec updates? Is this just a document in a repo? Maybe we can use [wei/pull](https://github.com/wei/pull) GitHub Action here as well.

If not, can you share more info, as I can work on it if you like @meowsbits.

meowsbits:
Hm.. this issue is an old and probably stale one. 

Since this issue was created we've added support for OpenRPC by reflection (which includes the spec'd methods). The dependabot PRs would have been useful if we wanted to update a static file (basically, the spec) that geth would serve at `rpc.discover`, but that's not really relevant or sensible anymore. 

If the spec changes and becomes incompatible with ours (since we meet it currently), then I would say it's up to the spec to bump their version number accordingly, and core-geth then until updating would just be implementing an _older_ version.

If we really want to be serious about assuring core-geth is aligned with the spec, then what comes to mind is a test that would `curl` the raw JSON spec and validate core-geth's `rpc.discover` result against it. Actually just typing this out loud starts to seem like a pretty decent test for the `rpc.discover` implementation anyways.

ziogaschr:
@meowsbits I think we can close this one based on #295

PR #295, opened 2021-01-12
node: add test for the discovery endpoint to validate OpenRPC compliance

ziogaschr:
Rel #15

I used https://github.com/xeipuuv/gojsonschema as I haven't found another similar package used already in geth.


node/node_test.go
```diff
@@ -244,6 +245,61 @@ func TestRegisterProtocols_OpenRPC_WS(t *testing.T) {
...
+	}
+	defer client.Close()
+
+	res := make(map[string]interface{})
+	err = client.Call(&res, "rpc_discover")
+	if err != nil {
+		t.Errorf("client call: %v", err)
+	}
+
+	openRPCSpecURL := "https://github.com/open-rpc/meta-schema/releases/download/1.14.0/open-rpc-meta-schema.json"
```
ziogaschr: @meowsbits do you know of a better URL to use so as we test always with the latest one. Using the master branch didn't worked.

node/node_test.go
```diff
@@ -244,6 +245,61 @@ func TestRegisterProtocols_OpenRPC_WS(t *testing.T) {
...
+
+	schemaLoader := gojsonschema.NewSchemaLoader()
+	schemaLoader.Draft = gojsonschema.Draft7 // force schema version
```
ziogaschr, line 284: I enforced using the Draft7 based on the OpenRPC spec, as the schema json doesn't set $schema as appropriate.

node/node_test.go
meowsbits: I looked at the tagged v1.14.0 version too, same issue I'm guessing: https://github.com/open-rpc/meta-schema/blob/1.14.0/schema.json#L21.

```json
      "enum": ["GENERATED FIELD: Do Not Edit - if you are seeing this unexpectedly, use released meta-schema (https://github.com/open-rpc/meta-schema/releases)"]
```

Looks like `/releases/` is the suggested way.

node/node_test.go
meowsbits: Noted. If this ever breaks we'll have to investigate more about the JSON Schema they're using served at https://meta.json-schema.tools/.

meowsbits: But LGTM.

node/node_test.go
```diff
@@ -244,6 +245,61 @@ func TestRegisterProtocols_OpenRPC_WS(t *testing.T) {
...
+	schemaLoader.AutoDetect = false
+	schema, err := schemaLoader.Compile(gojsonschema.NewReferenceLoader(openRPCSpecURL))
```
meowsbits: I'd be tempted to use a local copy of this... otherwise the tests depend on an internet connection (and the remote source existing unmutated)... :-S

meowsbits: (Rel comment above about URLs, too https://github.com/etclabscore/core-geth/pull/295/files#r555976156)

node/node_test.go
ziogaschr: We can leave it on auto, which will use hybrid mode based on [this](https://github.com/xeipuuv/gojsonschema#using-a-specific-draft):

> a hybrid mode is used which merges together functionality of all drafts into one mode

node/node_test.go
ziogaschr: Yup, same issue. I am afraid that we will forget to change it on a newer spec release. :(

Any ideas on how to handle this? I believe that they might have a URL somewhere serving the latest spec always.

node/node_test.go
ziogaschr: Oh, I thought I made this question on the description, but I probably forgot it. 
Yes I don't like either that we make the tests depend on the internet, though I liked the idea of pointing the URL to the latest spec version.

Let's bring the file locally and loaded from there

node/node_test.go
meowsbits: I think I'd just go with their suggested 
> use released meta-schema (https://github.com/open-rpc/meta-schema/releases)

And would maybe 

```
curl https://github.com/open-rpc/meta-schema/releases/download/1.14.0/open-rpc-meta-schema.json -o ./testdata/open-rpc-meta-schema-v1.14.0.json
```

and use

```go
 	openRPCSpecURL := "./testdata/open-rpc-meta-schema-v1.14.0.json" 
```

Thoughts?

node/node_test.go
meowsbits: Maybe we could add a CI-only test that cross-references the remote document with the local one? (The CI can be expected always have internet...)

ziogaschr: Solved in https://github.com/etclabscore/core-geth/pull/295/commits/9912e4c24875818472045b503ff67bbd6c1cc924

Do you prefer the file placed somewhere else? Or maybe within a **testdata** folder

node/node_test.go
ziogaschr: I had already done before reading the comment. But based on this I also moved it in a testdata folder.
Solved by https://github.com/etclabscore/core-geth/pull/295/commits/86157cb7fb658bc8c2c1d1d50f45e3ce578fbfdc

node/node_test.go
ziogaschr: @meowsbits do you mean in the style of a cron GitHub Action or a go test marked for CI only? Do you want to do this now, or open an issue for later?

meowsbits: I think we can separate it into it's own issue; IMO now this is doing exactly what it set out to do ✅. 
It might be nice to _be able_ to run the internet-requiring cross-referencing (:sweat_smile:) test locally too... maybe something like

```go
var openRPCSpecLatestTestEnvVarName := "COREGETH_SOME_GOOD_NAME"
func TestOpenRPCMetaSchemaSpecLatest(t *testing.T) {
    if os.Getenv("CI") == "" && os.Getenv(openRPCSpecLatestTestEnvVarName) {
        t.Skip()
    }
    // .. test that we're up-to-date and equivalent with remote canon
}
```

ziogaschr: actually, I am searching for things to work on, so let me tackle the test tomorrow in this PR. :)

meowsbits (approved):

ziogaschr:
@meowsbits please have a review of this https://github.com/etclabscore/core-geth/pull/295/commits/d0dae58b0f6bb833bd326a1cf56b0589806fbe53 and let's merge it

node/node_test.go
```diff
@@ -244,6 +245,83 @@ func TestRegisterProtocols_OpenRPC_WS(t *testing.T) {
...
+	openRPCSpecFile := "file://./testdata/open-rpc-meta-schema-1.14.0.json"
+	if err := discoverCompliesOpenRPCSpecTestRunner(openRPCSpecFile); err != nil {
+		t.Fatal(err)
+	}
+}
+
+var openRPCSpecLatestTestEnvVarName = "COREGETH_OPENRPC_LATEST_SPEC_ENDPOINT"
+
+// TestOpenRPCMetaSchemaSpecLatest tests compliance of discovery endpoint with the OpenRPC latest spec loaded remotelly as set in the CI
+func TestOpenRPCMetaSchemaSpecLatest(t *testing.T) {
```
meowsbits, line 311: - [ ] `COREGETH_OPENRPC_LATEST_SPEC_ENDPOINT` <- this still needs to be installed at the CI environment, or?
- [x] should this test check either or both?
  + equivalence of remote and local schemas
  + equivalence of remote and local version numbers

ziogaschr: Yes it has to be added at the CI, I had this in mind for after the merge.

Hmm, nice idea to also have something that checks if the local testdata file needs to be updated. Though at this point, that there is no endpoint to point on the latest/master JSON file, we are in control of both endpoints, which will probably point at the same location.

Now that I have written the above, I thought that I understood your intention wrongly. Do you want this test, to test using an external endpoint, or to compare the external latest JSON with the local one?

meowsbits: :) Originally I had the 
> compare the external latest JSON with the local one

in mind (since theoretically it should be exactly equivalent as using the local copy given the pegged version), but if there's a way to use the _latest_ remote version (instead of always replicating the local copy) then that seems like it would be valuable too.

ziogaschr: Solved by https://github.com/etclabscore/core-geth/pull/295/commits/386e26eb9eb96d713b9e43bb2474f11f32c8edf2

@meowsbits I added a test that compares the local file with the latest **released** one.
I removed the test that checks for discovery compliance with remote spec, as I prefer more that we are aware of the failing test (changed spec) and we fix it, instead of ignoring the local test, because the remote one passes.

```shell
COREGETH_OPENRPC_LATEST_SPEC_ENDPOINT=https://github.com/open-rpc/meta-schema/releases/latest/download/open-rpc-meta-schema.json CI=true go test -run ^TestOpenRPCMetaSchemaSpecLatest$ github.com/ethereum/go-ethereum/node
```

meowsbits: Yeap, nice. LGTM.

node/node_test.go
```diff
@@ -244,6 +246,104 @@ func TestRegisterProtocols_OpenRPC_WS(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to load local spec JSON file: %v", err)
+	}
+
+	var localSpec interface{}
+	if err := json.Unmarshal([]byte(localJSONFile), &localSpec); err != nil {
+		t.Fatalf("failed to unmarshal local JSON spec: %v", err)
+	}
+
+	if !reflect.DeepEqual(latestSpec, localSpec) {
```
meowsbits, line 342: Wait, does `reflect.DeepEqual(interface{}{}, interface{}{})` actually provide a good equivalence?

meowsbits: Nevermind; experimented with it a little (removing and modifying the local testdata/ file) and running the test; seems ✅. TIL about `reflect.DeepEqual` :joy:

meowsbits (approved):
LGTM

ziogaschr:
I will merge it. Can you please add this @meowsbits?

```
COREGETH_OPENRPC_LATEST_SPEC_ENDPOINT=https://github.com/open-rpc/meta-schema/releases/latest/download/open-rpc-meta-schema.json
```

PR #300, opened 2021-01-13
Improve existing docs

ziogaschr:
Suggestions:
- [x] Clean up the project README file and move sections in the docs. README will have a brief and will point to the docs page
- [x] ~~Examine which pages have to link at Ethereum pages and which we want to bring internally~~ (leaving it for another PR)

@meowsbits & @iquidus please comment on the above or directly edit the desc so as we create an initial list that I can work on


ziogaschr:
~~As it's not deployed, feel free to run it locally and check how it looks like~~

Actually it's deployed live

docs/apis/jsonrpc-apis.md
```diff
@@ -39,10 +42,11 @@ You'll need to use your own programming environments' capabilities (libraries, t
...
-
-**Note: Please understand the security implications of opening up an HTTP/WS based
-transport before doing so! Hackers on the internet are actively trying to subvert
-Ethereum nodes with exposed APIs! Further, all browser tabs can access locally
-running web servers, so malicious web pages could try to subvert locally available
-APIs!**
-
+Here you can check the available [JSON-RPC calls](https://playground.open-rpc.org/?schemaUrl={{ openrpc_latest_schema_endpoint }}){: target=_blank }.
+
+!!! attention
```
meowsbits: Is `!!! ` something special?

docs/apis/jsonrpc-apis.md
```diff
@@ -39,10 +42,11 @@ You'll need to use your own programming environments' capabilities (libraries, t
...
+    Please understand the security implications of opening up an HTTP/WS based
+    transport before doing so! Hackers on the internet are actively trying to subvert
+    Ethereum nodes with exposed APIs! Further, all browser tabs can access locally
+    running web servers, so malicious web pages could try to subvert locally available
+    APIs!**
```
meowsbits:
```suggestion
    APIs!
```

docs/apis/jsonrpc-apis.md
```diff
@@ -39,10 +42,11 @@ You'll need to use your own programming environments' capabilities (libraries, t
...
 need to speak [JSON-RPC](https://www.jsonrpc.org/specification) on all transports. You
 can reuse the same connection for multiple requests!
-
-**Note: Please understand the security implications of opening up an HTTP/WS based
-transport before doing so! Hackers on the internet are actively trying to subvert
-Ethereum nodes with exposed APIs! Further, all browser tabs can access locally
-running web servers, so malicious web pages could try to subvert locally available
-APIs!**
-
+Here you can check the available [JSON-RPC calls](https://playground.open-rpc.org/?schemaUrl={{ openrpc_latest_schema_endpoint }}){: target=_blank }.
```
meowsbits: This is broken for me as Github renders it

docs/core/alltools.md
```diff
@@ -5,14 +5,14 @@ hide:
...
-| `abigen`      | Source code generator to convert Ethereum contract definitions into easy to use, compile-time type-safe Go packages. It operates on plain [Ethereum contract ABIs](https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI) with expanded functionality if the contract bytecode is also available. However, it also accepts Solidity source files, making development much more streamlined. Please see our [Native DApps](https://github.com/ethereum/go-ethereum/wiki/Native-DApps:-Go-bindings-to-Ethereum-contracts) wiki page for details. |
+| **`geth`**    | Our main Ethereum Classic CLI client. It is the entry point into the Ethereum network (main-, test- or private net), capable of running as a full node (default),**** archive node (retaining all historical state) or a light node (retrieving data live). It can be used by other processes as a gateway into the Ethereum network via JSON RPC endpoints exposed on top of HTTP, WebSocket and/or IPC transports. `geth --help` and the [CLI page](../run-cli/#command-line-options) for command line options.          |
```
meowsbits:
```suggestion
| **`geth`**    | Our main Ethereum Classic CLI client. It is the entry point into the Ethereum network (main-, test- or private net), capable of running as a full node (default), archive node (retaining all historical state) or a light node (retrieving data live). It can be used by other processes as a gateway into the Ethereum network via JSON RPC endpoints exposed on top of HTTP, WebSocket and/or IPC transports. `geth --help` and the [CLI page](../run-cli/#command-line-options) for command line options.          |
```

docs/apis/openrpc.md
```diff
@@ -8,37 +8,43 @@ hide:
...
+            },
+            "summary": "current client version"
+          },
+
+    [...]
+    ```
 
-[...]
-```
+!!! Tip Better represantation of the discovery result at the OpenRPC playground
```
meowsbits:
```suggestion
!!! Tip Better representation of the discovery result at the OpenRPC playground
```

docs/apis/jsonrpc-apis.md
ziogaschr: Yes, it creates tooltips like in the below image. Here is the [docs](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) for it.

image

docs/apis/jsonrpc-apis.md
ziogaschr: Good catch man. So using this I am able to set a config variable with the actual path. Though if we want (makes sense) to not break support for our users who read the docs folders or the MD files directly, then it doesn't worth using variables.

I vote for removing variables. Please let me know what you think.

meowsbits: I agree... it's a nice-to-have convenience, but we can do almost the same thing with careful typing and find-and-replace :joy:

docs/apis/jsonrpc-apis.md
meowsbits: That looks really nice, but my concern is that it's not generic (or Github-flavored) markdown, which I'd generally like to stick with if possible for portability's sake. 

Can we instead approximate this rendering with something like

> __:warning: Attention__
> Please understand the security...

meowsbits: They do look really nice though... maybe I could be convinced...

meowsbits:
@ziogaschr Rel https://github.com/etclabscore/core-geth/pull/300#issuecomment-759706847
How is it deployed live? (eg. https://etclabscore.github.io/core-geth/core/run-cli/)

Is this because the Workflow triggers are _OR_ not _AND_? (_edited, I had it backwards) (So it runs on matching path change OR matching branch?) https://github.com/etclabscore/core-geth/blob/master/.github/workflows/docs-deploy.yml#L4-L20

ziogaschr:
> Is this because the Workflow triggers are AND not OR? (So it runs on matching path change OR matching branch?) https://github.com/etclabscore/core-geth/blob/master/.github/workflows/docs-deploy.yml#L4-L20

Yes I was thinking that this might be the cause.

docs/apis/jsonrpc-apis.md
ziogaschr: Resolved by https://github.com/etclabscore/core-geth/pull/300/commits/ab98c44c5d9e3032fa40047f1e3cf3bde1f4827b

docs/apis/jsonrpc-apis.md
ziogaschr: I really like them and I can't decide.
If we suggest everyone to read the docs from within the site, and based on that `!!! <attention>` for devs reading the docs directly the word next to `!!!` describes the level/intention of the message below, we might want to consider using it.

On the other hand, what you suggest above is a good alternative as well

ziogaschr: Your alternative serves our scope as well, but the material one is very nice 👍 

image

ziogaschr:
What about cleaning up the repo README file?

docs/apis/jsonrpc-apis.md
ziogaschr: We decided to continue with `!!! <attention>` way. If needed to be replaced later we can easily find all occurrences and replace them.

meowsbits:
I think cleaning up README is OK (for dedupe sake, at least ;)), but I'd suggest make it it's own PR.

ziogaschr:
Oops. @meowsbits I haven't seen that you prefer README changes in its own PR and have already committed them here. :(

ziogaschr:
@meowsbits and @iquidus this is ready for merge. Please have a look if you like.

meowsbits:
@ziogaschr No worries, LGTM! I just push a last commit https://github.com/etclabscore/core-geth/pull/300/commits/99a069ba41db6156a6f07ec54eebdb8fd9e93a43 adding some quick links for documentation sub-pages.

image

meowsbits (approved):
:page_facing_up: 🐿

PR #310, opened 2021-01-21
eth: update default extraData

iquidus:
updates the default extraData field, from geth -> CoreGeth to be consistent with both `geth version` and ethstats.


meowsbits (approved):

meowsbits (commented):
Would be good to have a test for this... just to prove at least that the encoding works as expected, say, for the new capitalized characters...

_EDIT_ Added ✅

PR #322, opened 2021-02-22
eth: add stale check before allowing re-enable of AF by sync status

meowsbits:
Mordor showed a scenario where despite a stale head,
sync achievement would re-enable AF features.

This causes the sync-status reenablement to
respect the stale head condition.

Date: 2021-02-22 06:19:13-06:00

### DIFF
--- a/eth/sync.go
+++ b/eth/sync.go
@@ -309,8 +309,13 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
 	op := peerToSyncOp(mode, peer)
 	if op.td.Cmp(ourTD) <= 0 {
 		// Enable artificial finality if parameters if should.
+		// - In full sync mode.
 		if op.mode == downloader.FullSync &&
+			// - Have enough peers.
 			cs.pm.peers.Len() >= minArtificialFinalityPeers &&
+			// - Head is not stale.
+			!(time.Since(time.Unix(int64(cs.pm.blockchain.CurrentHeader().Time), 0)) > artificialFinalitySafetyInterval) &&
+			// - AF is disabled (so we should reenable).
 			!cs.pm.blockchain.IsArtificialFinalityEnabled() {
 			cs.pm.blockchain.EnableArtificialFinality(true, "reason", "synced", "peers", cs.pm.peers.Len())
 		}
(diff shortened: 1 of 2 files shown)


ziogaschr:
The fix LGTM and is reasonable. I tested it and it worked fine.

I was only thinking if we want to disable MESS when chain becomes stale [here](https://github.com/etclabscore/core-geth/pull/322/files#diff-0da7dd4724e5bc1d3441ded068c9b182c67ca19bc4f67db5ba5d7f9906ea96edR293).

meowsbits:
> I was only thinking if we want to disable MESS when chain becomes stale here.

Following the link, I'm not sure I see exactly what you're pointing to... ? (Not getting a line number highlight for some reason.)

ziogaschr:
Here is an updated [link](https://github.com/etclabscore/core-geth/blob/902e56e85ed5faefd5ea854181f2c8c77b9b51aa/eth/sync.go#L293)

meowsbits:
Thanks, that one works :) 

Correct me if I'm wrong, but I think in that case the outcome is that MESS will only be _disabled_ (because of low peers). So there it doesn't care about chain head staleness; where staleness is 
- a condition for disabling (handled in the async loop [here](https://github.com/etclabscore/core-geth/blob/902e56e85ed5faefd5ea854181f2c8c77b9b51aa/eth/sync.go#L66)), and
- a condition for _not-enabling_, [here](https://github.com/etclabscore/core-geth/blob/902e56e85ed5faefd5ea854181f2c8c77b9b51aa/eth/sync.go#L317)

So this PR fixes the issue that _enabling_ MESS was not respecting/demanding _both_ head-staleness and peer-count conditions (that should have prevented it from being re-enabled after head staleness disabled it). 

Now written, this seems like a long way of saying:
- both conditions need to be met to enable
- thus, only one condition needs to fail to disable

ziogaschr:
LGTM, wasn't able to find the below condition on first check.

> a condition for disabling (handled in the async loop [here](https://github.com/etclabscore/core-geth/blob/902e56e85ed5faefd5ea854181f2c8c77b9b51aa/eth/sync.go#L66)),

ziogaschr:
Not related probably, the travis-ci tests fail because of a timeout (free tier limit?), can you confirm that. The failing test runs fine on my computer.

ziogaschr (approved):

meowsbits:
Yeap, checked that too. Same for me, fine on my computer.

PR #323, opened 2021-02-23
cmd/faucet: update Twitter auth/API changes to v1.1 + v2

meowsbits:
6ada9f0f38d14941ab02144319dba9482da18da5

This adds support for v1.1 and v2 in Twitter API.

See https://github.com/ethereum/go-ethereum/pull/22107

Date: 2021-02-23 05:53:57-06:00


ziogaschr (approved):
LGTM. Also tested the deployed version and it works nice. 

Only question is if we can/want to drop support for Twitter v1 API.

meowsbits:
Yea, roger that. Definitely a valid question. I haven't looked at the current state of Twitter's API; all I know is that I need `-twitter.token=xyx` for the thing to work :joy_cat:

Issue #321, opened 2021-02-21
no full syncked nodes in params/bootnodes_classic.go, Check nodes


lxgn:
Hi there,

In this list params/bootnodes_classic.go only one worked not syncked node.
more 30 not accessible.
enode://b9e893ea9cb4537f4fed154233005ae61b441cd0ecd980136138c304fefac194c25a16b73dac05fc66a4198d0c15dd0f33af99b411882c68a019dfa6bb703b9d@18.130.93.66:30303

I'm start sync from 0 block and my node long time wait and connect to ethereum (no classic) nodes with geth 1.9.xx

Check bootnodes list and change to normal worked with ethereum classic

For general questions please use [discord](https://discord.gg/nthXNEv) or the Ethereum stack exchange at https://ethereum.stackexchange.com.

#### System information

Geth version: `1.11.21`
OS & Version: Linux
Commit hash : (if `develop`)

#### Backtrace

````
[backtrace]
````

meowsbits:
Thanks for the report. 

These bootnodes seem to be relatively constant source of annoyance (and are slightly challenging to test and manage effectively because of peer capacities, availability rates, etc.)... which is why they're essentially deprecated in favor of the default [DNS peering](https://github.com/etclabscore/discv4-dns-lists/tree/etccore/all.classic.blockd.info) system.  

I'll look into a review of them too, and at the very least plan to add a note about their suspect utility and limitations.

meowsbits:
Closing per merge of #324, thanks again for noting this @lxgn.

PR #324, opened 2021-03-01
Feat/audit bootnodes ci

meowsbits:
Rel #321

Installs a Github Actions CI Workflow that audits the bootnodes listed as chain configuration defaults (for all chains).

It uses `./build/bin/devp2p discv4 ping`, pinging each enode. Nodes that respond pass, nodes that fail to respond... well, they fail. _You had one job...._

The workflow is configured to run
- as a cron job, once daily
- on pull requests that edit `params/bootnodes*go`
- or manually (via `workflow_dispatch`)

For `schedule` (cron) and `workflow_dispatch` triggers, the workflow will remove failing enodes and create a pull request with that change set.

For `pull_request` triggers, the action will _not_ create a PR; it will just just fail and the light will turn red :x:.

As for the actual outcome of auditing the bootnodes, a complementary PR is here: https://github.com/meowsbits/core-geth/pull/15.

### DIFF
--- a/.github/workflows/audit-bootnodes.yml
+++ b/.github/workflows/audit-bootnodes.yml
@@ -61,10 +61,11 @@ jobs:
               fi
             done
           done < <(find ./params -type f -name '*bootnode*go')
+          test $status -eq 0 || git --no-pager diff
           exit $status
 
       - name: Create Pull Request
-        if: github.event_name == 'schedule'
+        if: github.event_name != 'pull_request'
         uses: peter-evans/create-pull-request@v3
         with:
           commit-message: 'params: remove unresponsive bootnodes'
--- a/.github/workflows/audit-bootnodes.yml
+++ b/.github/workflows/audit-bootnodes.yml
@@ -73,5 +73,5 @@ jobs:
           body: |
             These bootnodes did not respond to `./build/bin/devp2p discv4 ping` requests.
 
-            For details: ${{GITHUB_SERVER_URL}}/${{GITHUB_REPOSITORY}}/actions/runs/${{GITHUB_RUN_ID}}.
+            For details: ${{github.github_server_url}}/${{github.github_repository}}/actions/runs/${{github.github_run_id}}.
           branch: remove-stale-bootnodes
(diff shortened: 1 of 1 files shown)


.github/workflows/audit-bootnodes.yml
```diff
@@ -0,0 +1,77 @@
...
+on:
+
+  schedule:
+    - cron: '0 6 * * *'
+
+  pull_request:
+    branches:
+      - 'master'
+    paths:
+      - 'params/bootnode*'
```
meowsbits, line 12: My intention here is for the workflow to trigger with a `pull_request` on `master` AND where a file matching the pattern under `paths` is modified. (Not OR.)

meowsbits: I've run some tests to experiment and prove that my logic about the AND vs. OR is correct, and it appears to be: https://github.com/meowsbits/ci-tests/actions.

.github/workflows/audit-bootnodes.yml
```diff
@@ -0,0 +1,77 @@
...
+          go build -o build/bin/devp2p ./cmd/devp2p
+
+      - name: Audit Bootnodes
+        shell: bash
+        run: |
+          healthcheck() {
+            enode_list="$(cat "$1" |grep -E 'enode://' | sed 's-"--g' | sed 's-\t--g'| sed 's/,.*//g')"
+            echo "$enode_list" | while read -r line; do
+              if >/dev/null 2>&1 ./build/bin/devp2p discv4 ping "${line}"; then
+                >&2 echo "PASS: ${line}"
```
ziogaschr, line 45: Woow, nice use of redirection. I am not good at bash and I am learning new stuff by checking your code @meowsbits  :)

ziogaschr (approved):
LGTM, nice use of GitHub actions to check the bootnodes. I run it and it seems we have a lot of failing bootnodes at the moment. Will we clean them in a separate PR?

meowsbits:
> Will we clean them in a separate PR?

Yea, that's what I was thinking.

PR #329, opened 2021-03-08
EVMC v7

### LINKED ISSUE #282: evmc: v7
Core-geth's [./evmc](https://github.com/etclabscore/core-geth/tree/master/evmc) adoption and implementation is currently pegged at the latest release version of the v6 major series (v6.3.1).

As noted at the [EVMC releases page](https://github.com/ethereum/evmc/releases), v7+ introduces support for the Istanbul EVM upgrades. As suggested by its major bump, it establishes a breaking change to the ABI along with quite a few other non-backward-compatible changes (see https://github.com/ethereum/evmc/releases/tag/v7.0.0). 

This issue proposes providing support for the v7 series of EVMC.

Related
- PR: #57 Feat/evmc6
- Issue: #55 Support EVMC6
- https://github.com/ledgerwatch/turbo-geth/blob/85219b651e3f3b7156ab2250bc6fbcc9a9a5ea0a/core/vm/evmc.go

meowsbits:
Resolves #282


tests/state_test.go
```diff
@@ -50,12 +51,10 @@ func TestState(t *testing.T) {
...
 	if *testEVM != "" {
-
-		// These fail because these forks were not supported at this version of the EVMOne .so.
-		// The EVMOne version (0.2.0) is the latest EVMC v6 compatible version.
-		st.skipFork("Constantinople") // Only support
-		st.skipFork("Istanbul")
-		st.skipFork("ETC_Phoenix")
+		// These interpreters fail Constantinople (but pass ConstantinopleFix).
+		if strings.Contains(*testEVM, "aleth") || strings.Contains(*testEVM, "evmone") {
+			st.skipFork("^Constantinople$")
```
meowsbits, line 56: I haven't looked at why these fail (nor at if they're even expected to pass) (the symptom is the typical `mismatched post state root`). 

Since Constantinople was (and is) followed by `ConstantinopleFix` (aka `Petersbug`) I'm inclined to consider the `Constantinople` fork as an ugly duckling and allow it to fail here without further pursuit. From its disuse (at least relative to ETH and ETC), I think it's probably fair to consider it "poorly supported"... at least as far as the the EVM domain goes (and from our current time onward). 

Note further that the subsequent `Istanbul` and its `ETC_Phoenix` variant run without complaint.

iquidus: I'm ok with this line of thinking. Other than the initial foundation testnet, no networks should be activating Constantinople and petersburg separately.

ziogaschr:
Nice work @meowsbits. I reviewed the code diff and I will try to give it a run tomorrow.

meowsbits:
@ziogaschr Am I remembering right that you mentioned `hera` wasn't building for you on OSX? Is this possible to address here? (Even if only by skipping the test on OSX if it can't easily be supported?)

ziogaschr (approved):
LGTM. 

With regard failing to run `make test-evmc` in MacOS. I examined this and it has to do with the case that lib files for Linux end at `.so` while on MacOS at `.dylib`. You can see this even in the used [libs instructions](https://github.com/ewasm/hera#building-hera).

I believe that it is ok to keep out testing of the bindings of external EVMs in MacOS is fine.

meowsbits:
@ziogaschr What do you think about adding conditions to the `build/hera.sh`, `build/ssvm.sh`, etc. scripts that assert that the OS is Linux (as the only currently supported OS by those scripts), and exit immediately with a `"This is script is only currently configured to work on Linux. Please see ${provider} documentation for instructions to build in other environments."` ?

At least that way we can give a more helpful and immediate exit (rather than risking developers having to fall down a rabbit hole debugging).

ziogaschr:
@meowsbits it's fixed on #351. Please check it and merge it in this PR if you like it.

meowsbits:
53e345fa88f400b9590545bc0f86d2703f634559 resolves merge conflicts in go.mod.

Issue #283, opened 2021-01-11
evm: cross-client benchmarks


meowsbits:
I have a hypothesis that core-geth's EVM implementation isn't optimized relative to it's peers turbo-geth and go-ethereum. Reasons for this include:
- core-geth removes the `EVM.chainRules` field in favor of directly accessing the `EVM.chainConfig`. AFAIK `chainRules` is basically a cache, so removing it should yield a less efficient implementation.
- core-geth's jump table is created with each EVM, rather than assigning one from a small set of available chainconfig/fork instruction sets. Though more extensible, this is probably not as efficient.

This issue proposes developing and installing a programmatic comparison of these cross-client benchmarks so we can see whether or not the above hypothesis holds water.

PR #336, opened 2021-03-12
ci: comparative benchmarks for vm, core, and trie operations

meowsbits:
Rel #283

Provides 
- new benchmarking functionality for the JSON VM tests suite (`tests/vm_bench_test.go`)
- 3 new Github Actions workflows that compare go-ethereum vs. core-geth benchmarks for
  + the JSON VM tests,
  + `core/`, and
  + `trie/` packages.

Intention is to provide a reliable means of establishing efficiency and optimization comparison with upstream source.
Example test runs can be seen on my personal fork at https://github.com/meowsbits/core-geth/actions. 

The actions' workflows run go-ethereum and core-geth jobs in parallel, and compare the results using `benchstat`.

### DIFF
--- a/.github/workflows/bench-core.yml
+++ b/.github/workflows/bench-core.yml
@@ -1,0 +1,96 @@
+name: Benchmark Core
+on:
+  push:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  bench_core_geth:
+    name: Benchmark core-geth
+    runs-on: ubuntu-latest
+    timeout-minutes: 360
+    steps:
+      - name: Set up Go 1.x
+        id: go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.16
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Benchmark etclabscore/core-geth
+        id: bench
+        run: |
+          go test -short ./core -count 1 -p 1 -timeout 60m -run NONE -bench=. -v |& tee core-geth.txt
+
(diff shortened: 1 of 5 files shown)


.github/workflows/bench-core.yml
```diff
@@ -0,0 +1,96 @@
...
+        id: go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.16
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          repository: 'ethereum/go-ethereum'
+          ref: 'v1.9.25'
```
ziogaschr, line 47: It's important to not forget to update the version here from now on. Wondering if we can fetch this version from somewhere in another step and use it here.

ziogaschr (approved):
Nice work man. It's useful to be able to keep an eye on such stats.

.github/workflows/bench-core.yml
meowsbits: Yea, this crossed my mind too. The issue with dynamic updates would be that there are no guarantees that benchmark tests will stay constant across upstream versions (although in practice, this is probably an edge case); and I wanted to maintain consistency for the test control group. My thinking was that when we start to actually use these, we _should_ expect ourselves to have the basic states of the test groups in mind when analyzing their results.

I'm not hard and fast on hardcoding it, but that was my reasoning. If you'd like to push for dynamic versioning I would be more than happy to see what you have in mind.

meowsbits:
Just rebased on master@f0dd598bf86747d87ca9691fe668631cc18ae138.

Issue #334, opened 2021-03-12
rpc api: eth_getBlockByNumber "pending" block handling of nullable fields


meowsbits:
#### System information

OS & Version: Linux
Commit hash : a7e6418fa4b13e7b0b98ef15d4149f41fd1739fe

#### Expected behaviour

I expect the documentation, go-ethereum, and core-geth to align in their
description and implementation of `eth_getBlockByNumber("pending", [true|false])`. 

The ETH docs I'm looking at (and citing) are here: https://eth.wiki/json-rpc/API.
CoreGeth docs are here: https://etclabscore.github.io/core-geth/JSON-RPC-API/modules/eth/#eth_getblockbynumber.

Rel #332 

#### Actual behaviour

Currently I believe there is no complete agreement (neither between docs nor clients). 
I will attempt to describe the divergences.

- `nonce`
  + Should be `null`. This value depends on the PoW solution, which is obviously not yet available.
  + __Docs:__ Should be `NULL`.
  + __core-geth:__ Is omitted. _Incorrect_.
  + __go-ethereum:__ Is `null`. OK.

- `mixHash`
  + Should be `null` or zero-value. This value depends on the PoW solution, which is obviously not yet available.
  + __Docs:__ Missing entirely.
  + __core-geth:__ Zero-value. OK.
  + __go-ethereum:__ Filled. _Incorrect_.

- `number`
  + Should not be `null`. Should be "latest" + 1.
  + __Docs:__ Should be `NULL`. _Incorrect_. 
  + __core-geth:__ Is filled. OK.
  + __go-ethereum:__ Is filled. OK.

- `hash`
  + Should be `null`. Block cannot be hashed until all other fields are finalized.
  + __Docs:__ Should be `NULL`. OK. 
  + __core-geth:__ Is omitted. _Incorrect_.
  + __go-ethereum:__ Is `null`. OK.

- `miner`
  + Should be `null`. Miner is not yet known.
  + __Docs:__ Should be filled. _Incorrect_. 
  + __core-geth:__ Is omitted. _Incorrect_.
  + __go-ethereum:__ Is `null`. OK.

- `totalDifficulty`
[...]

PR #341, opened 2021-03-15
Fix/eth_getBlockByNumber and eth_getUncle... nullable fields

meowsbits:
Rel #334

This is working toward implementing and testing `eth_getBlockByNumber` for `pending` blocks achieving complete parity with ethereum/go-ethereum (... and with reference to the docs, which seem to be occasionally a source of noise rather than truth).

One question remaining concerns the `totalDifficulty` field. Currently (as tested) both core-geth and go-ethereum return `null` for the `pending` block. This field is _not present_ in the docs (cited at above ref'd issue). Logic at the API level determines the existence of this field conditionally on an `inclTxes` boolean. I'm remembering wrangling with this a while ago, but need to do some more digging to get the details again.


meowsbits:
Ref https://github.com/etclabscore/ethereum.go-ethereum/pull/11/files. This branch shows that `ethclient.GetHeader(-1)`, which should return the `pending` block, in fact does not.  This is a bug in the `ethclient` implementation (although with roots potentially at `types.Header` JSON Unmarshal). Note that this branch and the demonstrated bug is at ethereum/go-ethereum. 

Though related, _this_ issue is first concerned with the RPC service (ie server) providing valid information (rather than the Go `ethclient` implementation of a client consuming this data properly).

internal/ethapi/api.go
```diff
@@ -1198,12 +1198,7 @@ func NewRPCMarshalHeaderTFromHeader(header *types.Header) *RPCMarshalHeaderT {
 // rpcMarshalHeaderTSetTotalDifficulty sets the total difficulty field for RPC response headers.
 // If the hash is unavailable (ie in Pending state), the value will be 0.
 func (s *PublicBlockChainAPI) rpcMarshalHeaderTSetTotalDifficulty(ctx context.Context, header *RPCMarshalHeaderT) {
-	hash := header.Hash
-	if hash == nil {
-		c := common.Hash{}
-		hash = &c
-	}
-	header.TotalDifficulty = (*hexutil.Big)(s.b.GetTd(ctx, *hash))
+	header.TotalDifficulty = (*hexutil.Big)(s.b.GetTd(ctx, *header.Hash))
```
meowsbits, line 1201: This now matches the ethereum/go-ethereum implementation.
> https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1187-L1199

meowsbits:
Re: `totalDifficulty`.
- https://github.com/ethereum/go-ethereum/pull/20460
- https://github.com/ethereum/go-ethereum/issues/19024#issuecomment-505358363

I've set it to behave equivalently with ethereum/go-ethereum. This means that it will be included (potentially `null`... under certain conditions...). This might spin off into it's own issue.

meowsbits:
This commit https://github.com/etclabscore/core-geth/pull/341/commits/67ca5fcf0398beba097b5ba27517d8079ad7ea58 fixes this issue: https://github.com/etclabscore/core-geth/issues/344

meowsbits:
There was a `panic` at the CI. https://travis-ci.org/github/etclabscore/core-geth/jobs/763301644

At first glance it seems unrelated to this change set, but has now happened twice in a row (after 1 re-run). Will look into it tomorrow.

ziogaschr (approved):
LGTM

PR #343, opened 2021-03-17
.github/workflows: fix bash condition syntax for showing git diff

meowsbits:
Fixes 'integer expression expected' error

Date: 2021-03-17 07:14:20-05:00


meowsbits:
Merging as trivial.

PR #345, opened 2021-03-17
Implement stateDiff tracer, compatible with OpenEthereum

ziogaschr:
This PR adds **stateDiff** tracer that is compatible with OpenEthereum.

> **stateDiff State difference**. Provides information detailing all altered portions of the Ethereum state made due to the execution of the transaction.

The **stateDiff** tracer can be used in the following **Ad-hoc Tracing** methods:

- [x] trace_call
- [x] trace_callMany
- [ ] trace_rawTransaction
- [ ] trace_replayBlockTransactions
- [ ] trace_replayTransaction

p.s.: The remaining unchecked **Ad-hoc Tracing** methods will be added in a new PR, and will have stateDiff tracer support out of the box.

### Differences with OpenEthereum

1. SSTORE in some edge cases are being persisted in state but are being removed from stateDiff storage results. (Happens only on 2 transactions on Mordor testnet)
2. When error ErrInsufficientFundsForTransfer happens, OpenEthereum leaves the tracer run producing negative balances, though using safe math for overflows it returns 0 balance, on the other hand the **to** account receives the full amount. Core-geth removes only the gas cost from the **sender** and adds it to the **coinbase** balance.
3. Same as in 2, but on top of that, the **sender** account doesn't have to pay for the gas cost even. In this case, core-geth returns an empty JSON, as in reality this transaction will remain in the tx_pool and never be executed, neither change the state.
4. On OpenEthereum the **block gasLimit** is set to be [`U256::max()`](https://github.com/openethereum/openethereum/blob/main/crates/ethcore/src/client/client.rs#L1857), which leads into problems on contracts using it for pseudo-randomness. On core-geth, we believe that the user utilising the `trace_*` wants to see what will happen in reality, though we leave the block untouched to its true values.
5. When an internal call fails with out of gas, and its state is not being persisted, we don't add it in stateDiff output, as it happens on OpenEthereum.


ziogaschr:
@meowsbits where's the best place to add the **Differences with OpenEthereum** section from the above notes in our docs? Do you remember if we have a way to add them within the generated docs?

eth/api_tracer.go
```diff
@@ -824,16 +914,65 @@ func (api *PrivateDebugAPI) TraceCall(ctx context.Context, args ethapi.CallArgs,
...
 		if err != nil {
 			return nil, err
 		}
 	}
 
 	// Execute the trace
-	msg := args.ToMessage(api.eth.APIBackend.RPCGasCap())
-	vmctx := core.NewEVMContext(msg, header, api.eth.blockchain, nil)
-	return api.traceTx(ctx, msg, vmctx, statedb, config)
+	var results = make([]interface{}, len(txs))
```
ziogaschr, line 924: The below code is exact copy of the one in `TraceCall` which can be decoupled on its own function. I kept it here until we merge from go-ethereum and we take the refactor where they moved the trace stuff into a package.
I will refactor the logic then on a new PR.

ziogaschr: Also the state_diff tests will need this logic. At the moment, we have a 3rd partial copy there.

eth/tracers/internal/tracers/prestate_tracer.js
```diff
@@ -18,7 +18,8 @@
 // the transaction from a custom assembled genesisT block.
 {
 	// prestate is the genesisT that we're building.
-	prestate: null,
+	prestate: {},
```
ziogaschr, line 21: This tracer, is being used to get the prestate for tests to be used as genesis. I changed this as well to utilise the CapturePreEVM as well.

ziogaschr:
Please don't review it. I will refactor it on top of #348 once it's finished.

ethclient/ethclient_test.go
```diff
@@ -809,7 +809,8 @@ var allRPCMethods = []string{
 	"personal_unlockAccount",
 	"personal_unpair",
 	"trace_block",
-	"trace_filter",
```
meowsbits, line 812: Why is `trace_filter` removed?

ziogaschr: Because it is subscription based.

ziogaschr:
We decided to include this in the next release and refactor for #348 after that.

eth/api_tracer_parity.go
```diff
@@ -71,6 +71,21 @@ func setConfigTracerToParity(config *TraceConfig) *TraceConfig {
 	return config
 }
 
+func enchanceParityOutput(res interface{}, config *TraceConfig) interface{} {
```
meowsbits: Typo here `s/enchance/enhance`. Maybe `decorateParityOutput` instead?

eth/api_tracer_parity.go
```diff
@@ -71,6 +71,21 @@ func setConfigTracerToParity(config *TraceConfig) *TraceConfig {
...
+	if config.ParityCompatibleOutput {
```
meowsbits: For this logic, I would recommend moving this condition out of the function (eg. to line 252). 

```
if shouldDecorate {
    decorate()
}
```

vs. 

```
decorate() {
    if !shouldDecorate {
        return
    }
}

File: eth/api_tracer.go [hunk 6]
```diff
@@ -60,9 +60,10 @@ const (
 // TraceConfig holds extra parameters to trace functions.
 type TraceConfig struct {
 	*vm.LogConfig
-	Tracer  *string
-	Timeout *string
-	Reexec  *uint64
+	Tracer                 *string
+	Timeout                *string
+	Reexec                 *uint64
+	ParityCompatibleOutput bool
```
meowsbits: Does `TraceConfig` have custom JSON Un/Marshal methods? Where/how are these fields configured? (Forgive my naivete... :blush: )

ziogaschr: No, it doesn't. It just uses the default field names as the name. They just fallback to the default value for the type at the moment, or if no key is set, then the whole config is nil.

I was thinking to make this explicit when I joined ETC and started working on Tracers, but left it like this to not trigger any future conflicts with go-ethereum.

ziogaschr: Related, here I initiate it using the default if needed. https://github.com/etclabscore/core-geth/blob/feat/trace-state-diff/eth/api_tracer_parity.go#L66

eth/api_tracer_parity.go
ziogaschr: Thanks man, handled at 3582231. You triggered me to add support for all available JS tracers as well. 👍

eth/tracers/tracer.go
```diff
@@ -561,6 +568,25 @@ func wrapError(context string, err error) error {
 	return fmt.Errorf("%v    in server-side tracer function '%v'", err, context)
 }
 
+func (jst *Tracer) CapturePreEVM(env *vm.EVM, inputs map[string]interface{}) error {
```
meowsbits, line 573: Other tracer `Capture` methods are defined as interface methods in core/vm/logger.go's `Tracer` type. This interface is implemented by `JSONLogger`, `StructLogger`, and `Tracer`. Should `CapturePreEVM` be likewise defined in that interface? Why or why not?

docs/JSON-RPC-API/trace-module-overview.md
```diff
@@ -0,0 +1,53 @@
...
+These APIs allow you to get a full externality trace on any transaction executed throughout the blockchain.
+
+- [x] trace_block
+- [x] trace_transaction
+- [x] trace_filter (doesn't support address filtering yet)
+- [ ] trace_get
+
+### Available tracers
+
+- `callTracerParity` Transaction trace.
```
meowsbits:
```suggestion
- `callTracerParity` Transaction trace returning a response equivalent to OpenEthereum's (aka Parity) response schema. For documentation on this response value see _______.
```

docs/JSON-RPC-API/trace-module-overview.md
```diff
@@ -0,0 +1,53 @@
+# "trace" Module Overview
+
+The trace module is for getting a deeper insight into transaction processing. It includes two sets of calls; the transaction trace filtering API and the ad-hoc tracing API. You can find the documentation for the supported methods [here](/JSON-RPC-API/modules/trace/).
```
meowsbits, line 3: Some context here regarding the relationship of methods in `trace_` which may be similar to ones in `debug_` would, I think, be helpful.

docs/JSON-RPC-API/trace-module-overview.md
```diff
@@ -0,0 +1,53 @@
...
+        "tracer": "stateDiffTracer",
+        "timeout: "10s",
+        "reexec: "10000",               // number of block to reexec back for calculating state
+        "parityCompatibleOutput": true  // for Ad-hoc Tracing only in order the response is nested as in OpenEthereum output
+    }
+    ```
+
+## "stateDiff" tracer differences with OpenEthereum
+
+1. **SSTORE** in some edge cases persists data in state but are not being returned on stateDiff storage results on OpenEthereum output. *(Happens only on 2 transactions on Mordor testnet)*.
```
meowsbits: Citations of canonical chain data should probably include state information, like _as of block 1,500,214_.

core/vm/interpreter.go
```diff
@@ -254,8 +254,7 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 		// consume the gas and return an error if not enough gas is available.
 		// cost is explicitly set so that the capture state defer method can get the proper cost
 		if operation.dynamicGas != nil {
-			var dynamicCost uint64
-			dynamicCost, err = operation.dynamicGas(in.evm, contract, stack, mem, memorySize)
+			dynamicCost, err := operation.dynamicGas(in.evm, contract, stack, mem, memorySize)
```
meowsbits: This will assign two new variables instead of only one (reassigning the return value `err`). Is this just a pretty-fication change?

eth/api_tracer.go
meowsbits: Got it. For ease of reference

```go
// setConfigTracerToParity forces the Tracer to the Parity one
func setConfigTracerToParity(config *TraceConfig) *TraceConfig {
	if config == nil {
		config = &TraceConfig{}
	}

	if config.Tracer == nil {
		tracer := "callTracerParity"
		config.Tracer = &tracer
	}

	return config
}
```

Used as 

```go
// Block returns the structured logs created during the execution of
// EVM and returns them as a JSON object.
// The correct name will be TraceBlockByNumber, though we want to be compatible with Parity trace module.
func (api *PrivateTraceAPI) Block(ctx context.Context, number rpc.BlockNumber, config *TraceConfig) ([]interface{}, error) {
	// Fetch the block that we want to trace
	var block *types.Block

	switch number {
	case rpc.PendingBlockNumber:
		block = api.eth.miner.PendingBlock()
	case rpc.LatestBlockNumber:
		block = api.eth.blockchain.CurrentBlock()
	default:
		block = api.eth.blockchain.GetBlockByNumber(uint64(number))
	}
	// Trace the block if it was found
	if block == nil {
		return nil, fmt.Errorf("block #%d not found", number)
	}

	config = setConfigTracerToParity(config)

	traceResults, err := traceBlockByNumber(ctx, api.eth, number, config)
	if err != nil {
		return nil, err
	}
```

In this case it seems to me like the `setConfigTracerToParity` is actually _defaulting_ the config to Parity in case the value is nil (as opposed to `forces the Tracer to the Parity one`), or? Maybe just a rename for that function?

meowsbits: I'm curious about the JSON field encoding here. Normally JSON fields for exported methods will use capitalized field names;

```go
//cat main.go
package main

import (
        "encoding/json"
        "fmt"
)

type MyStruct struct {
        Reexec             *string
        ParityCompatOutput bool
}

func main() {
        ex := &MyStruct{}

        b, _ := json.MarshalIndent(ex, "", "    ")
        fmt.Println(string(b))
}
```

```sh
go run main.go
{
    "Reexec": null,
    "ParityCompatOutput": false
}
```

And in the internal/web3ext.go it doesn't look like this arg is getting a custom formatter. (~~Also, should `callMany` have same formatters as `call`?~~ No, because args are different; callMany uses an array of CallArgs. Should this have its own formatter?).  

```js
		new web3._extend.Method({
			name: 'call',
			call: 'trace_call',
			params: 3,
			inputFormatter: [web3._extend.formatters.inputCallFormatter, null, null]
		}),
		new web3._extend.Method({
			name: 'callMany',
			call: 'trace_callMany',
			params: 3,
			inputFormatter: [null, null, null]
		}),
```

eth/api_tracer.go
```diff
@@ -140,7 +142,7 @@ func (api *PrivateDebugAPI) TraceChain(ctx context.Context, start, end rpc.Block
 // traceChain configures a new tracer according to the provided configuration, and
 // executes all the transactions contained within. The return value will be one item
 // per transaction, dependent on the requested tracer.
-func (api *PrivateDebugAPI) traceChain(ctx context.Context, start, end *types.Block, config *TraceConfig) (*rpc.Subscription, error) {
+func traceChain(ctx context.Context, eth *Ethereum, start, end *types.Block, config *TraceConfig) (*rpc.Subscription, error) {
```
meowsbits, line 145: What do you think about adding some commentary (maybe even package- or file-level) that would explain this extracted method and signature change. Future merges and and conflict-resolving humans :wink: will probably come across this more than once, and I would be nice to be able to get a little refresher on the reasoning and design behind it.

eth/api_tracer_parity.go
```diff
@@ -56,11 +66,30 @@ func setConfigTracerToParity(config *TraceConfig) *TraceConfig {
 		config = &TraceConfig{}
 	}
 
-	tracer := "callTracerParity"
-	config.Tracer = &tracer
+	if config.Tracer == nil {
```
meowsbits, line 69: Rel my other comment at https://github.com/etclabscore/core-geth/pull/345#discussion_r604327234

eth/api_tracer_parity.go
```diff
@@ -56,11 +66,30 @@ func setConfigTracerToParity(config *TraceConfig) *TraceConfig {
...
+		tracer := "callTracerParity"
+		config.Tracer = &tracer
+	}
+
 	return config
 }
 
+func decorateParityOutput(res interface{}, config *TraceConfig) interface{} {
```
meowsbits:
```suggestion
// decorateParityOutput formats trace results the way Parity does.
// Docs: https://openethereum.github.io/JSONRPC-trace-module
// Example:
/*
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "output": "0x",
    "stateDiff": null,
    "trace": [{
      "action": { ... },
      "result": {
        "gasUsed": "0x0",
        "output": "0x"
      },
      "subtraces": 0,
      "traceAddress": [],
      "type": "call"
    }],
    "vmTrace": null
  }
}
/*
func decorateParityOutput(res interface{}, config *TraceConfig) interface{} {
```

eth/tracers/tracer.go
```diff
@@ -238,6 +238,13 @@ func (dw *dbWrapper) pushObject(vm *duktape.Context) {
 		return 1
 	})
 	vm.PutPropString(obj, "exists")
+
+	// Push the wrapper for statedb.Empty
+	vm.PushGoFunction(func(ctx *duktape.Context) int {
+		ctx.PushBoolean(dw.db.Empty(common.BytesToAddress(popSlice(ctx))))
+		return 1
+	})
+	vm.PutPropString(obj, "isEmpty")
```
meowsbits: I think `isEmpty` is a better name, but it looks like other wrapper use the snakeCased method name verbatim. Worth changing to `"empty"` just to adhere to the pattern? At most this is a nitpick...

internal/web3ext/web3ext.go
```diff
@@ -796,6 +796,18 @@ web3._extend({
...
+			name: 'call',
+			call: 'trace_call',
+			params: 3,
+			inputFormatter: [web3._extend.formatters.inputCallFormatter, null, null]
+		}),
+		new web3._extend.Method({
+			name: 'callMany',
+			call: 'trace_callMany',
+			params: 3,
+			inputFormatter: [null, null, null]
```
meowsbits: `callMany` uses an array of CallArgs. Should this have its own formatter?  Rel https://github.com/etclabscore/core-geth/pull/345#discussion_r604330305

eth/api_tracer_parity.go
meowsbits: Added another touch to it - https://github.com/etclabscore/core-geth/pull/345/commits/a5d00e081c2427a26778bd7f66789da6bab8ce6a

Per the commit message, up for your review too

ziogaschr: @meowsbits I am fine with keeping the a5d00e08, though I left it as it was as I am afraid of future dev mistakes that can lead to a bug by not changing the if condition in all places. Reason is that this exact call/check will be called from 5 (and maybe even 8) different places/functions once all the trace methods are being supported.

Thinking it twice and based on where you placed the check now, by doing a slight modification on `setConfigTracerToParity` we can change the condition to be `if config.ParityCompatibleOutput {` which makes much more sense then and it's harder to break the logic on future. Will check if it makes sense.

eth/api_tracer.go
ziogaschr: > In this case it seems to me like the setConfigTracerToParity is actually defaulting the config to Parity in case the value is nil (as opposed to forces the Tracer to the Parity one), or? Maybe just a rename for that function?

Good catch. It's a comment issue. It has been left off, from the initial tracing work. Will be updated.

> I'm curious about the JSON field encoding here. Normally JSON fields for exported methods will use capitalised field names;

True. Currently this is only used to Unmarshal JSON and the default parser handles `camelCase` as well. I followed go-ethereum practice without paying much attention to this one and you are right that this can create issues later on.
Do you think it's good to handle this now? and if yes, shall we do this in this or another PR?

> And in the internal/web3ext.go it doesn't look like this arg is getting a custom formatter. (Also, should callMany have same formatters as call? No, because args are different; callMany uses an array of CallArgs. Should this have its own formatter?).

Yes, this is a great idea. Though we have to check out how the `internal/jsre/deps/web3.js` is being packaged externally, and if such change will be easy to maintain (conflicts resolution) later on.
@meowsbits idea to see if we can extend this functionality in `web3ext.go` is great. I will examine this, as I have some concerns with regards the custom function context.

eth/tracers/tracer.go
ziogaschr: Initially, I left it out of the interface as it will only be used for the JS tracers and not the rest.
On the other hand is a good addition, and later on rest tracers might have some benefits using it.
So bbc4a7b adds `CapturePreEVM` in the interface, to make everything cleaner.

Thanks @meowsbits.

core/vm/interpreter.go
ziogaschr: Reverted 53eaf17. Great catch, as it will trigger issues.

eth/api_tracer.go
ziogaschr: No need to do this now, as very soon, after the #348 merge, this separation will be removed.  🎉

eth/tracers/tracer.go
ziogaschr: No strong preference. `isEmpty` feels to better serve its purpose. Changed to `empty` at 16fb0ec.

docs/JSON-RPC-API/trace-module-overview.md
ziogaschr: Done at 2e486f6

docs/JSON-RPC-API/trace-module-overview.md
ziogaschr: Done at 2e486f6

docs/JSON-RPC-API/trace-module-overview.md
ziogaschr: Done at 2e486f6

internal/web3ext/web3ext.go
ziogaschr: Resolved with e0b8ae0

eth/api_tracer_parity.go
ziogaschr: Resolved with 84bf302

eth/api_tracer.go
ziogaschr: Resolved by e0b8ae0 and 84bf302.

eth/api_tracer_parity.go
ziogaschr: Resolved by 84bf302

ziogaschr:
@meowsbits I handled all your comments. Please have a look at them and re-review. Let me know if you find anything else as based on your comments we did some good hardening.

eth/api_tracer_parity.go
```diff
@@ -74,11 +74,34 @@ func setConfigTracerToParity(config *TraceConfig) *TraceConfig {
 	return config
 }
 
-func decorateParityOutput(res interface{}, config *TraceConfig) interface{} {
+// decorateResponse applies formatting to trace results if needed.
+func decorateResponse(res interface{}, config *TraceConfig) (interface{}, error) {
```
meowsbits, line 78: :+1:

eth/api_tracer.go
meowsbits: Nicely done with https://github.com/etclabscore/core-geth/pull/345/commits/e0b8ae0ce39231bf76229dbbf5142aa496bdd139!

meowsbits: > I followed go-ethereum practice without paying much attention to this one and you are right that this can create issues later on. Do you think it's good to handle this now? and if yes, shall we do this in this or another PR?

Understood. As long as there are tests ensuring that all the fields of the Config get decoded as expected, then I think it's fine as-is.

ethclient/ethclient_test.go
meowsbits: I'm pretty sure there's a way to keep this in the docs/OpenRPC generation. 

`eth_subscribe` establishes subscriptions with a different signature pattern (eg. `eth_subscribe ["newHeads"]` -> `(*api.Eth) SubscribeNewHeads(...) (rpc.Subscription, error)`.
So if this convoluted pattern can do it, then `trace_filter` must be able to too :wink:. 

It might require some noodling in node/openrpc.go.

eth/tracers/internal/tracers/prestate_tracer.js
```diff
@@ -70,13 +74,12 @@
...
-		if (this.prestate === null){
-			this.prestate = {};
-			// Balance will potentially be wrong here, since this will include the value
-			// sent along with the message. We fix that in 'result()'.
+		if (!this.hasStepInited) {
+			this.hasStepInited = true;
+
 			this.lookupAccount(log.contract.getAddress(), db);
 		}
+
```
meowsbits, line 82: Man this is a lot tidier. Well done!

eth/tracers/testdata/state_diff_tracer_big_slow.json
```diff
@@ -0,0 +1,100 @@
...
+    "timestamp": "1577401639",
+    "alloc": {
+      "0x877bd459c9b7d8576b44e59e09d076c25946f443": {
+        "balance": "0x17b561689d13557dfc3d",
+        "nonce": "7653",
+        "code": "0x",
+        "storage": {}
+      }
+    },
+    "config": {
```
meowsbits, line 22: A note that using the Mordor config (both configuration format and configuration itself) for test cases limits their applicability to just core-geth; ethereumgo-ethereum won't be able to run these. 

Is it possible to use `&goethereum.ChainConfig` schemas here so they could be more easily shared upstream? And to use an arbitrary chain configuration/genesis that ethereum/go-ethereum would be able to support?

eth/tracers/tracers_test.go
```diff
@@ -473,3 +474,161 @@ func BenchmarkCallTracerParity(b *testing.B) {
...
+	// as the Transaction is being run on top of the block transactions,
+	// which might lead into ErrInsufficientFundsForTransfer error
+	canTransfer := func(db vm.StateDB, sender common.Address, amount *big.Int) bool {
+		if msg.From() == sender {
+			return true
+		}
+		return core.CanTransfer(db, sender, amount)
+	}
+
+	// If the actual transaction would fail, then their is no reason to actually transfer any balance at all
```
meowsbits:
```suggestion
	// If the actual transaction would fail, then there is no reason to actually transfer any balance at all
```

eth/tracers/tracers_test.go
```diff
@@ -473,3 +474,161 @@ func BenchmarkCallTracerParity(b *testing.B) {
...
+		Coinbase:    test.Context.Miner,
+		BlockNumber: new(big.Int).SetUint64(uint64(test.Context.Number)),
+		Time:        new(big.Int).SetUint64(uint64(test.Context.Time)),
+		Difficulty:  (*big.Int)(test.Context.Difficulty),
+		GasLimit:    uint64(test.Context.GasLimit),
+		GasPrice:    msg.GasPrice(),
+	}
+	_, statedb := tests.MakePreState(rawdb.NewMemoryDatabase(), test.Genesis.Alloc, false)
+
+	// Store the truth on wether from acount has enough balance for context usage
```
meowsbits:
```suggestion
	// Store the truth on whether from account has enough balance for context usage
```

ethclient/ethclient_test.go
ziogaschr: That's great. I will have a look at it. 
Do you think it worths handling this in a next PR, so as we merge this one? As we have to add the remaining bits for `trace_fitler` as well in a next PR.

eth/api_tracer.go
```diff
@@ -802,19 +811,100 @@ func (api *PrivateDebugAPI) TraceTransaction(ctx context.Context, hash common.Ha
...
+		}
+		if block == nil {
+			return nil, fmt.Errorf("block %v not found: %v", blockNrOrHash, err)
+		}
+		// try to recompute the state
+		reexec := defaultTraceReexec
+		if config != nil && config.Reexec != nil {
+			reexec = *config.Reexec
+		}
+		_, _, statedb, err = computeTxEnv(eth, block, len(block.Transactions())-1, reexec)
```
meowsbits, line 837: IIRC, this `len(block.txs)-1` (vs. `0`) was related to a bug found at ethereum/go-ethereum w/r/t the computed env, or? Just out of curiousity, did report that or they get it fixed?

eth/tracers/testdata/state_diff_tracer_big_slow.json
ziogaschr: Great point. Will have a look at this together with the rest stuff discussed yesterday. 
I believe this fits nicely on a new PR and doesn't worth blocking this PR at the moment. 
I agree, that it has to be done, before making an upstream PR.

eth/tracers/internal/tracers/state_diff_tracer.js
```diff
@@ -0,0 +1,537 @@
...
+				continue;
+			}
+		}
+	},
+
+	// includeOpError checks for specific VM OP errors
+	includeOpError: function(err) {
+		return err
+			&& err.indexOf('contract address collision') > -1;
+			// && (err.indexOf('insufficient balance for transfer') > -1
```
meowsbits: Dead code or developer clue?

eth/api_tracer.go
ziogaschr: Yes, that's indeed a bug that we handled here. From what I recall they handled this fine in the trace package refactor, but I will have to check for sure.

It's on my list of things to refer/PR upstream.

eth/tracers/internal/tracers/state_diff_tracer.js
```diff
@@ -0,0 +1,537 @@
...
+			fromAcc._error = false;
+
+			// don't update from state anymore, as we applied our customs calcs
+			fromAcc._final = true;
+		}
+
+		if (this.stateDiff[toAccountHex] !== undefined) {
+			var toAcc = this.stateDiff[toAccountHex];
+
+			// TODO: check if it needs to be added back
```
meowsbits: :tomato: A real pending TODO or DONE?

eth/tracers/internal/tracers/state_diff_tracer.js
ziogaschr: It's a developer clue on a flow I found on OE code, but I am not sure if reachable on state_diff and I couldn't find a use case or proof that it is needed.

I will remove it as it doesn't look good. Either way, I have some notes in my agenda with regards such things.

eth/tracers/internal/tracers/state_diff_tracer.js
ziogaschr: Leftover. Has been tested and it can be removed

ziogaschr: Resolved by 523cafe

eth/tracers/internal/tracers/state_diff_tracer.js
ziogaschr: Resolved by f553df4

eth/tracers/testdata/state_diff_tracer_big_slow.json
meowsbits: SGTM!

meowsbits (approved):
Awesome work.

iquidus (approved):

eth/api_tracer.go
ziogaschr: @meowsbits this has been resolved upstream.

ethclient/ethclient_test.go
ziogaschr: @meowsbits can you please let me know if this is what you had in mind? #345

eth/tracers/testdata/state_diff_tracer_big_slow.json
ziogaschr: Resolved by https://github.com/etclabscore/core-geth/pull/355/commits/208da5589f626f9200ec11f557b373dfc5645e40

PR #353, opened 2021-04-01
<- Merge/foundation release/1.10.1 resolved: cross-client tests and tests generators/fillers

meowsbits:
This is a refactoring and improving of cross-client test generation.

Issues provoking this
- Parity configuration (does not support Belin EIPs); I have `deprecated` Parity config-format support for EIPs installed after Istanbul. 
    + Remove Makefile commands that do Parity stuff (but core functionality remains, just not conveniently through Makefile).

Does
- Documents configurations used in State tests at `tests/testdata/GeneralStateTests_configs.json` as a map keyed on the subtest Fork names used in those tests, eg `Berlin` and `ETC_Magneto`.
    + These configurations are now core-geth format instead of parity format.

I'll make more notes inline with the code to point out important stuff.

See generated tests diff at https://github.com/etclabscore/tests/pull/4


Makefile
```diff
@@ -102,15 +107,17 @@ tests-generate: tests-generate-state tests-generate-difficulty ## Generate all t
 tests-generate-state: ## Generate state tests.
 	@echo "Generating state tests."
 	env COREGETH_TESTS_GENERATE_STATE_TESTS=on \
+	env COREGETH_TESTS_CHAINCONFIG_FEATURE_EQUIVALENCE_COREGETH=on \
```
meowsbits, line 110: `COREGETH_TESTS_CHAINCONFIG_FEATURE_EQUIVALENCE_COREGETH` forces all `Forks` values used by the tests to be `confp.Convert`ed on `init` to `&coregeth.ChainConfig` type values. These config values are dumped JSON encoded to an adjacent file. 

No longer do clients' test runner have to somehow magically know what "Berlin" means; they can look it up and see exactly what config should be used.

params/config_classic.go
```diff
@@ -60,6 +60,8 @@ var (
 		EIP145FBlock:  big.NewInt(9573000),
 		EIP1014FBlock: big.NewInt(9573000),
 		EIP1052FBlock: big.NewInt(9573000),
+		// EIP1283FBlock:   big.NewInt(9573000),
+		// PetersburgBlock: big.NewInt(9573000),
```
meowsbits, line 64: I've just added these as documentation to reflect the related Constantinople config; there were issues with EIP1234 that came up in other places and this was part of my investigation.

params/types/coregeth/chain_config_configurator.go
```diff
@@ -600,6 +603,12 @@ func (c *CoreGethChainConfig) SetEthashEIP649Transition(n *uint64) error {
...
 	}
 
+	if eip1234 := c.GetEthashEIP1234Transition(); eip1234 != nil {
+		if *eip1234 <= *n {
+			return nil
+		}
+	}
+
 	c.ensureExistingRewardSchedule()
 	c.BlockRewardSchedule[*n] = vars.EIP649FBlockReward
```
meowsbits, line 614: These two changes are important. The prevent a configuration from overwriting an already-configured EIP1234 value with a same-block EIP649 configuration; essentially codifying that EIP649 can not be set after EIP1234.

params/types/multigeth/multigethv0_chain_config_configurator.go
```diff
@@ -313,6 +313,8 @@ func (c *ChainConfig) SetEIP1052Transition(n *uint64) error {
 }
 
 func (c *ChainConfig) GetEIP1283Transition() *uint64 {
+	// This is special case handling for multigeth configuration.
+	// It makes the tests pass.
 	if c.ConstantinopleBlock != nil && c.PetersburgBlock != nil {
 		if c.ConstantinopleBlock.Cmp(c.PetersburgBlock) == 0 {
 			return nil
```
meowsbits, line 320: This is not an ideal thing, and it differs from the `goethereum` implementation. This is due, in part, to core-geth's original inheritance of the multi-geth configuration format and need to support some old strange versions of the multi-geth config.

tests/state_test_util.go
```diff
@@ -85,18 +85,17 @@ type stInfo struct {
 	LLLCVersion        string                 `json:"lllcversion"`
 	Source             string                 `json:"source"`
 	SourceHash         string                 `json:"sourceHash"`
-	WrittenWith        string                 `json:"written-with"`
-	Parent             string                 `json:"parent"`
-	ParentSha1Sum      string                 `json:"parentSha1Sum"`
```
meowsbits, line 90: These three fields were specific to core-geth test generation metadata. I've removed them since their information be inferred with version control, and I use the `FilledWith` field to define the core-geth version (instead of `WrittenWith`).

tests/state_test_util.go
```diff
@@ -85,18 +85,17 @@ type stInfo struct {
...
-	Chainspecs         chainspecRefsT         `json:"chainspecs,omitempty"`
 	Labels             map[string]interface{} `json:"labels,omitempty"`
```
meowsbits, line 88: `Labels` was apparently added at the tests suite.

tests/state_test_util.go
meowsbits: Also `Chainspecs` is disused since the tests generation creates the `GeneralStateTests_configs.json` which defines the test configurations; making references to configs by file unnecessary.

params/types/multigeth/multigethv0_chain_config_configurator.go
meowsbits: I looked at generalizing the interaction of `EIP1283` and `EIP1283Disable` as `Disable`-type methods at the `Configurator.(Equivalent|Compatible)`, but decided that was too involved for now; and this case currently only applies to 1283, being the only EIP which has been explicitly disabled.

PR #356, opened 2021-04-12
Test for OpenRPC subscription methods on trace/debug modules

ziogaschr:
Resolves https://github.com/etclabscore/core-geth/pull/345#discussion_r605569982


meowsbits:
Thanks for getting this going! 
Have you tried generating docs with this and see if they are as expected? I haven't yet, but have a feeling there's going to be a mismatch; please correct me if I'm wrong.

Pub/Sub is not handled well automatically by the reflector; it's a "special" case for the way ethereum/go-ethereum does (or does not) do method reflection and mapping.

For example, `eth_subscribe("newHeads")` uses the method `SubscribeNewHeads` (or is it `NewHeads`? can't remember exactly); does `trace_filter` work the same (or use `(api *TraceAPI) Filter(...)`)? 

The current hack-around for `eth_subscribe`'s weirdness is the `RPCSubscription` mocks in node/openrpc.go, where the subscription's basically-_enum_ values (eg. `newHeads`, `newSideHeads`, `newPendingTransactions`...) are provided in hardcode as type-mappings (see `rpcSubscriptionParamsNameD`). 

I think we may need to either provide a similar kind of mock pattern for `debug_` and `trace_` modules (as you have started), or get a little fancier with additional conditions to [the catch-all pub/sub skip logic](https://github.com/etclabscore/core-geth/pull/356/files#diff-555b8d6d556def392f76b30e25597eed281d38cc3992572c54137a9b641fb1eaR145).

TL;DR: I don't think `debug_subscribe` and `trace_subscribe` exist.

```go
	        // Exclude methods that handle subscriptions, but do so without adhering to the conventional code pattern.
		// Eg. *filters.PublicFiltersAPI.SubscribeNewHeads handles eth_subscribe("newHeads"), but there
		// isn't a method called `eth_subscribeNewHeads`. So we blacklist all these methods and use
		// the mock subscription receiver type RPCSubscription.
		// This pattern matches all strings that start with Subscribe and are suffixed with a non-zero
		// number of A-z characters.
		if isPubSub(method.Type) {
			// This catches MOST of them (except SubscribeSyncStatus)
			return false
		}
```

ziogaschr:
No I haven't generated the docs. I will.
I think I see what you mean now. Will have a look myself and I will ping you.

Thanks man

node/openrpc.go
```diff
@@ -329,7 +369,8 @@ var blockNumberOrHashD = fmt.Sprintf(`{
           ]
         }`, blockNumberD, commonHashD, requireCanonicalD)
 
-var rpcSubscriptionParamsNameD = `{
+var rpcEthSubscriptionParamsNameD = `{
+		"title": "subscriptionName",
```
ziogaschr, line 374: @meowsbits added this, as it wasn't working for params without it even for `eth_subscribe`.

node/openrpc.go
```diff
@@ -208,25 +210,59 @@ func newOpenRPCDocument() *go_openrpc_reflect.Document {
 }
 
 /*
-The following struct RPCSubscription and RPCSubscription.Unsubscribe method
+The following structs *RPCSubscription and *RPCSubscription.Unsubscribe method
```
ziogaschr, line 214: @meowsbits I want to play a bit in order to find a better way of merging those 3 duplicated implementations where only the **subscriptionName** type in `subscribe` changes.

Let me know if you have an idea.

node/openrpc.go
```diff
@@ -339,6 +380,20 @@ var rpcSubscriptionParamsNameD = `{
 		]
 	}`
 
+var rpcDebugSubscriptionParamsNameD = `{
+		"title": "subscriptionName",
+		"oneOf": [
+			{"type": "string", "enum": ["traceFilter"], "description": "Returns traces for the filtered transactions within a range of blocks."}
```
ziogaschr: I still have to collect all the `debug_*` subscription based methods

ziogaschr:
### Generated docs sample

image

node/openrpc.go
meowsbits: Have you experimented at all with getting rid of or modifying the [general "skip PubSub" case](https://github.com/etclabscore/core-geth/blob/735fd8804dae5f612b7bddea3b441767784caa04/node/openrpc.go#L143) to be more permissive (even if only allowing `debug` and `trace` receivers through)?

ziogaschr: No, I haven't. I will give it a try. Thanks

ziogaschr: I have tried it here, which adds the `trace_filter` nicely, although it ~~can~~can't be used in the way documented, which leads to issues. What do you think? Did you mean something else @meowsbits?

image

meowsbits: X-linking: https://github.com/etclabscore/core-geth/pull/364#pullrequestreview-643261538

ziogaschr:
@meowsbits I added generating docs for both `*_subscribe`, but also more accurate logs for the subscriptionable methods like `trace_filter` (in this case, I change the examples section).

1. I still need to fix the examples...
2. I am thinking if it makes sense in this case to add a new field per method `Subscription: bool`

meowsbits:
Per [this comment at OpenRPC/spec](https://github.com/open-rpc/spec/issues/211#issuecomment-513509637) we could potentially use `x-subscription: true` (or similar) as a specification extension on the method annotations.

ziogaschr:
@meowsbits can you review it? To see what's done, check the `trace_subscribe` and `trace_filter` in trace.md file.

p.s.: At the moment I haven't enabled the same for the `eth` package, but we can do that there too. For example documenting the `eth_newHeads`.

> Per [this comment at OpenRPC/spec](https://github.com/open-rpc/spec/issues/211#issuecomment-513509637) we could potentially use `x-subscription: true` (or similar) as a specification extension on the method annotations.

This is a great addon, but not straightforward to implement. As discussed, let's keep this for later.

meowsbits (approved):
If this latest commit (https://github.com/etclabscore/core-geth/pull/356/commits/ce80270fa344ce41e6267e5e4a503604ea92be32) looks OK to you, everything LGTM! 

(I've left the methods commented because that lists serves a duty both as a test case, but also as a "behind-the-scenes" developer reference.)

Issue #358, opened 2021-04-14
EIP 3436 Clique Block Choice Rule


q9f:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3436.md

1. Choose the block with the most total difficulty.
2. Then choose the block with the lowest block number.
3. Then choose the block whose validator had the least recent in-turn block assignment.
4. Then choose the block with the lowest hash.

q9f:
https://ethereum-magicians.org/t/eip-3436-expanded-clique-block-choice-rule/5809

meowsbits:
> __The current specification of Clique allows for multiple competing blocks from producers but does not provide any strategies to pick blocks aside from the current "highest total difficulty" rule.__ This EIP proposes a four step choice rule of highest total difficulty, shortest chain, most recently in-turn, and lowest hash. This would prevent deadlocks that have occurred in production systems.

https://eips.ethereum.org/EIPS/eip-225: Clique proof-of-authority consensus protocol

meowsbits:
A screen shot taken during the latest deadlock on Kotti.

![2021-04-06-073308_1886x551_screenshot](https://user-images.githubusercontent.com/45600330/115024250-71156a80-9e85-11eb-9b96-542d97cfe049.png)

![image](https://user-images.githubusercontent.com/45600330/115026277-ebdf8500-9e87-11eb-8ff1-4312cf8eedc9.png)

PR #360, opened 2021-04-16
Feat/eip 3436 clique block priority rules

meowsbits:
Resolving #358.

### Notes

- EIP3436 is a "soft" feature; it doesn't need to be activated with a hard fork. Despite that, I've added it as a transitional value. 
- The configuration field for this value has been added to the `CliqueConfig` struct, adjacent to `Period` and `Epoch` values. This allows the Configuration interfaces to share the implementation quite easily.
- As noted in https://github.com/etclabscore/core-geth/pull/360/commits/64b0b7f9bc2ac4213d9de3abdbd571508f661167, the whole "should reorg" logic in `writeBlockWithState` is pretty awkward.


meowsbits:
Weird, it seems the commit on Github are listed out of order.

image

core/blockchain.go
```diff
@@ -1615,19 +1616,49 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
...
 	currentBlock = bc.CurrentBlock()
 	if !reorg && externTd.Cmp(localTd) == 0 {
 		// Split same-difficulty blocks by number, then preferentially select
 		// the block generated by the local miner as the canonical block.
 		if block.NumberU64() < currentBlock.NumberU64() {
-			reorg = true
-		} else if block.NumberU64() == currentBlock.NumberU64() {
+			reorg = true // EIP-3436 (2): Choose the block with the lowest block number.
+
+		} else if block.NumberU64() == currentBlock.NumberU64() && bc.Config().GetConsensusEngineType().IsEthash() {
```
meowsbits: Add a restriction to this scope limiting it to `Ethash`.

When this scope is applied to `Clique` (as originally), the always-false return value of `bc.shouldPreserve` yielded a truthy `reorg` value indescriminately 50% of the time.

core/blockchain_af.go
```diff
@@ -65,6 +66,9 @@ func (bc *BlockChain) EnableArtificialFinality(enable bool, logValues ...interfa
 // finality feature setting.
 // This status is agnostic of feature activation by chain configuration.
 func (bc *BlockChain) IsArtificialFinalityEnabled() bool {
+	if bc.Config().GetConsensusEngineType() != ctypes.ConsensusEngineT_Ethash {
+		return false
+	}
```
meowsbits, line 71: This should be a practical redundancy since AF is only reasonably applied to PoW consensus engines, but this enforces that. Questionably, this could cause silently unexpected operation if a user _does_ try to use an AF feature with a non-PoW protocol, since they won't receive an error or refusal on startup, but the features will simply be non-operational (implying that this is a logically half-assed change).

params/types/parity/parity.go
```diff
@@ -136,9 +137,9 @@ type ParityChainSpec struct {
 	Accounts map[common.UnprefixedAddress]*ParityChainSpecAccount `json:"accounts"`
 }
 
-func (c *ParityChainSpec) String() string {
+func (spec *ParityChainSpec) String() string {
 	cc := &ParityChainSpec{}
-	*cc = *c
+	*cc = *spec
```
meowsbits, line 142: Unrelated; a product of a global replacement for consistent receiver name `spec`.

consensus/consensus.go
```diff
@@ -115,6 +115,9 @@ type Engine interface {
 	// that a new block should have.
 	CalcDifficulty(chain ChainHeaderReader, time uint64, parent *types.Header) *big.Int
 
+	// Reorg defines block reorganization logic following consensus rules like heaviest chain.
```
meowsbits, line 118:
```suggestion
	// ElectCanonical defines block canon status logic following consensus rules like heaviest chain.
```

core/blockchain.go
```diff
@@ -208,7 +207,7 @@ type BlockChain struct {
 	vmConfig   vm.Config
 
 	badBlocks          *lru.Cache                     // Bad block cache
-	shouldPreserve     func(*types.Block) bool        // Function used to determine whether should preserve the given block.
+	shouldPreserve     func(*types.Header) bool       // Function used to determine whether should preserve the given block.
```
meowsbits, line 210: The `Block` data type is a superset of `Header`. The function assignments as implemented (and, afaik, intended) only need information from the Header. So this leans things up a little bit, and the refactor was very smooth across all uses.

ziogaschr:
@meowsbits do you have any ideas on running an actual test on this? 
Rather than only reviewing this PR's code and tests?

meowsbits:
Some thoughts only.

Ethash is infinitely more important than Clique; so testing that this doesn't break things for Ethash is the first priority.

I assume (from some empirical experience but no firm citations at the tip of my fingertips) that the decision making logic of canonical blocks should be well tested already, directly and/or indirectly. For example, the decision making for block import status should be intrinsically called with many of the `Blockchain`, `Miner`, and `eth` tests.

Basically I'm counting on the assumption that so much other (Ethash-based) stuff depends on this logic that if it were broken, a test _somewhere_ would have to break. :sweat: :woozy_face: 

Even if we allow this shaky hypothesis, the question of adding a test to show that the dealock cases the EIP claims to fix are actually resolved still remains... There are 2 scenarios provided in the Rationale section of EIP-3436.

meowsbits:
I'm working on writing tests for the two EIP3436 scenarios now. :memo:

paulgoleary:
We have ported these changes to our ETH fork here: https://github.com/qluvio/go-ethereum/tree/dev-port-clique-eip3436
We've been experiencing somewhat frequent stalls and as of today we've deployed these changes to all validators after ~1 month running on a parallel chain in our staging env.
We also don't have a definitive way to reproduce the issue, but with current use we've been seeing stalls on the order of once a month. I'll report back if we continue to see problems.

UPDATE: so far, so good but we *did* stub our toe on one issue. When the new code is in place but the config'd transition has not occurred yet. i.e. when this check is `false`:
```
c.config.EIP3436Transition == nil || c.config.EIP3436Transition.Cmp(current.Number) > 0
```
this causes the fork resolution rules to short-circuit after the block number and TD checks. However, the previous ETH rule also included a coin-flip as the final rule. Without that rule the new behavior seems to be definitively worse because no node will take *any* re-orgs so no forks get resolved at all.
We inadvertently did this and our chain was halting every other day until we figured it out.

meowsbits:
Closing this. Kotti (ETC's Clique-based network) has sunsetted. Without further movement here, Clique is a low priority and its nuanced consensus issues even more so.

PR #363, opened 2021-04-23
WIP: Feat/eip 3436 clique block priority rules [tests]

meowsbits:


consensus/clique/clique_test.go
```diff
@@ -113,3 +118,395 @@ func TestReimportMirroredState(t *testing.T) {
...
+					1A x       x
+					2B  x      y
+					3C   x      x
+					4D    x     y
+					5E     x     x
+					6F      x    y
+					7G       x-
+					8H        -
+
+				FIXME(meowsbits): This scenario yields a "recently signed" error
```
meowsbits: :eyes: PTAL

As noted in the comments here, I am wondering if the spec'd scenario is actually valid.
https://eips.ethereum.org/EIPS/eip-3436

consensus/clique/clique_test.go
```diff
@@ -113,3 +118,395 @@ func TestReimportMirroredState(t *testing.T) {
...
+				when attempting to import Signer 5 (really #6 b/c zero-indexing) into the
+				second fork.
+
+				On that fork, the sequence of signers is specified to be
+				... 7, 0, 1, 2, 3, 4, 5, 6, 1, 3, 5
```
meowsbits: I'm reading this as signer `5` should not actually be allowed to validate this block, since it's eg. `17-13=4` blocks from their last validation, where the limit should be `8/2+1=5`. This would mean that the `error recently signed` coming from the test for this scenario would be right.

meowsbits:
Is merged to #360

PR #364, opened 2021-04-23
WIP: Test for OpenRPC subscription methods on trace/debug modules / whitelist-style pubsub skips

meowsbits:
I know it's annoying that I've polluted the diff with the regenerated docs... :shrug: 

The issue this is trying to solve is that #356's https://github.com/etclabscore/core-geth/pull/356/files#diff-98749d04d0efbddfeb407114a7c48c0309646642e7ff67b807f39ecdc5809924R924 adds `debug_(un|)subscribe` and `trace_(un|)subscribe` which AFAIK don't exist; those modules don't use the same weird pattern that `eth_subscribe` does.

Please feel free to cherry-pick, ignore, modify, or whatever here. Just a sketch of concept.


ethclient/ethclient_test.go
```diff
@@ -952,12 +952,12 @@ var allRPCMethods = []string{
 	"eth_getBlockTransactionCountByHash",
 	"eth_getBlockTransactionCountByNumber",
 	"eth_getCode",
-	"eth_getFilterChanges",
-	"eth_getFilterLogs",
+	// "eth_getFilterChanges",
+	// "eth_getFilterLogs",
```
meowsbits, line 956: I've left the removed methods commented for :eyes:

node/openrpc.go
```diff
@@ -142,10 +140,15 @@ func newOpenRPCDocument() *go_openrpc_reflect.Document {
 		// the mock subscription receiver type RPCSubscription.
 		// This pattern matches all strings that start with Subscribe and are suffixed with a non-zero
 		// number of A-z characters.
-		if isPubSub(method.Type) {
-			// This catches MOST of them (except SubscribeSyncStatus)
-			return false
+		pkgStr := method.Type.In(0).String()
+		if (isPubSub(method.Type) && pkgStr == "*eth.PublicEthereumAPI") || pkgStr == "*filters.PublicFilterAPI" {
```
meowsbits, line 144: This is the important part.

I'm using the stringer condition because using `reflect.TypeOf(&eth.PublicEthereumAPI)` will cause an import cycle error.

I _think_ its correct that all the methods from `*filters.PublicFilterAPI` can be skipped since, AFAIU, these are all "unconventional" methods in service of either `eth` or `debug` or `trace` methods which use subs.

meowsbits: Related to https://github.com/etclabscore/core-geth/pull/356/files#r618692596

ziogaschr:
Moving our discussion here as well. `(debug|trace)_(un)subscribe` exists and this is the way to use the needed methods.
This approach looks cleaner and on top of this one we have to create a new template generator for the subscription methods.

meowsbits:
Oh man, well ~~if~~ since `(trace|debug)_(un|)subscribe` exists, then.... 
a.) I'm very wrong
b.) your approach might be way better?

ziogaschr:
Let me try something later tonight or tomorrow on this PR as I am working on trace stuff now. I think you opened my eyes one more time and the solution is in the intersection of both PRs. Thanks

meowsbits:
Closing since @ziogaschr has got an improved version already going on at #356 , which is this PR's base branch.

Issue #365, opened 2021-04-27
[Tracking] go-ethereum/1.10.2 merge TODOs and tidbits


meowsbits:
This issue is tracking checklists and TODOs related to the merge of ethereum/go-ethereum 1.10+ series releases and the associated protocol upgrade features. This release series includes features implemented in the Berlin fork, and proposed for Ethereum Classic's tentative Magneto fork.

### Overview

PRs are stacked.

- `master` <- go-ethereum/1.10.1 https://github.com/etclabscore/core-geth/pull/348
- https://github.com/etclabscore/core-geth/pull/348 <- go-ethereum/1.10.2 https://github.com/etclabscore/core-geth/pull/357

### TODOs

-  [x] `tests/testdata`.  This gets bumped. ethereum/tests has been refilled and now includes something like 150 new tests for Berlin features. I regenerated the tests based on the commit upstream is using.
    - [ ] These generated tests need reviewed. https://github.com/etclabscore/tests/pull/4
        - Read https://github.com/etclabscore/tests/pull/4#issue-607707514. Review footprint, configuration schemas and presence.
        - This is where the state tests are generated: https://github.com/etclabscore/core-geth/blob/merge/foundation-release/1.10.1-resolved/tests/state_mgen_test.go#L204. Please review.
        - This is where the difficulty tests are generated: https://github.com/etclabscore/core-geth/blob/merge/foundation-release/1.10.1-resolved/tests/difficulty_mgen_test.go#L141. Please review.
- [x] @meowsbits `core/blockchain_af_test.go`. I resolved conflicts from the hip there. Review. 
- [x] @meowsbits Make sure header `nil` checks/validations are in place. IIRC core-geth has this and go-ethereum does not. I believe must be in place since I believe there is a miner test testing nil header cases. 
- [x] #354 `panic: invalid number of list elements: 0`
- [x] @ziogaschr Review and handle `X_Override` values (eg. `BerlinOverride` and `--override.berlin`, etc.), eg. w/r/t `SetupGenesisBlockWithOverride` and default config. re #367
- [x] @meowsbits `--debug` is missing. On purpose? :( 
    _It becomes `log.debug`._ See a7443c0e5c for related fixes regarding deprecated `pprof` flags.
- [x] @meowsbits Add core-geth additional chain flags, re https://github.com/etclabscore/core-geth/pull/357#pullrequestreview-635112410. 
    Done with 49bd5e70385382528187bd42365392b4a91fad5f.
[...]

PR #373, opened 2021-05-10
Fix allowing of selecting protocol versions after merge of 1.10.x

ziogaschr:
Keep in mind that Eth protocol version 63 has been dropped.
_EDIT(meowsbits) (And 62 was dropped a long time ago.)_

Ref #365

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -172,7 +172,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	if bcVersion != nil {
 		dbVer = fmt.Sprintf("%d", *bcVersion)
 	}
-	log.Info("Initialising Ethereum protocol", "versions", config.ProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
+	log.Info("Initialising Ethereum protocol", "network", config.NetworkId, "dbversion", dbVer)
 
 	if !config.SkipBcVersionCheck {
 		if bcVersion != nil && *bcVersion > core.BlockChainVersion {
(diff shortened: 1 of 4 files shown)


eth/handler.go
```diff
@@ -51,6 +51,10 @@ var (
 	syncChallengeTimeout = 15 * time.Second // Time allowance for a node to reply to the sync progress challenge
 )
 
+// DefaultProtocolVersions are the supported versions of the `eth` protocol (first
+// is primary).
+var DefaultProtocolVersions = []uint{eth.ETH66, eth.ETH65, eth.ETH64}
```
ziogaschr, line 56: This had to be set in the eth package as the protocol has nested packages logic for (eth|snap) protocols.

eth/backend.go
```diff
@@ -543,7 +543,7 @@ func (s *Ethereum) BloomIndexer() *core.ChainIndexer   { return s.bloomIndexer }
 // Protocols returns all the currently configured
 // network protocols to start.
 func (s *Ethereum) Protocols() []p2p.Protocol {
-	protos := eth.MakeProtocols((*ethHandler)(s.handler), s.networkID, s.ethDialCandidates)
+	protos := eth.MakeProtocols((*ethHandler)(s.handler), s.networkID, s.config.ProtocolVersions, s.ethDialCandidates)
 	if s.config.SnapshotCache > 0 {
 		protos = append(protos, snap.MakeProtocols((*snapHandler)(s.handler), s.snapDialCandidates)...)
```
ziogaschr, line 548: At the moment we only have a flag for the eth protocols (66|65|64) and not for the snap protocols. I am not sure if we prefer handling the snap protocol version now or leave them out, as it's just one at the moment.

meowsbits: Noted, thanks for noticing this. SGTM to leave as-is (no flag), then think about adding one when there's more than one to choose from.

meowsbits:
So the `EthProtocols` flag is core-geth specific. (IIRC we added it as a kind of just-in-case option for security cases like Ethereum Classic's Phoenix upgrade where a subset of nodes (OpenEthereum, I think) failed to support 65 properly and nodes supporting (only) lower versions were important in "bridging" the network.) 

This patch fixes that flag to be effectual again, and to support the 66,65,64 set. 

Just making sure I understand the motivation and context here.

ziogaschr:
Yes, you recall correctly. This has been added on core-geth only for this exact reason.

meowsbits:
Note that I've created https://github.com/etclabscore/core-geth/pull/375 in response to the failing tests here; it's just an impatient timeout issue.

meowsbits (approved):

PR #375, opened 2021-05-11
cmd/geth,internal/cmdtest: add KillTimeout field to TestCmd struct

meowsbits:
TestGethStartupLogs was failing because of shutdown
timeout, via https://travis-ci.org/github/etclabscore/core-geth/jobs/770485427.

This allows a configurable process kill timeout window
and provides an unchanged default (5 seconds).

Note that the field isn't protected or really closely
integrated with the process runner at (ie via runGeth function),
so the kill field basically just gets set on the fly and
hopes that it gets set before the shutdown logic is called.

Date: 2021-05-11 05:56:29-05:00


ziogaschr (approved):
LGTM

PR #378, opened 2021-05-13
Merge/foundation release/1.10.3

ziogaschr:


core/blockchain.go
```diff
@@ -1231,74 +1216,19 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
...
-						}
-					}
-					if err := batch.Write(); err != nil {
-						return 0, err
-					}
-					batch.Reset()
-					deleted = deleted[0:]
+				return i, fmt.Errorf("containing header #%d [%x..] unknown", block.Number(), block.Hash().Bytes()[:4])
+			}
+			// TODO(ziogaschr): check removal of https://github.com/etclabscore/core-geth/commit/f71d78cb26aad5bc7e4df76fadd082b69ca2de82
```
ziogaschr: @meowsbits can you please check if you want this logic maintained? I think with the latest upstream changes it is not needed.

meowsbits: - `core: remove old conversion to shuffle leveldb blocks into ancients ` https://github.com/ethereum/go-ethereum/commit/83375b08731d95fdc91b6e1576db596c21384a3d
- https://github.com/ethereum/go-ethereum/pull/22739

LOKTM :heavy_check_mark:

ziogaschr: Resolved by 848cd11

eth/tracers/api_test.go
```diff
@@ -362,19 +362,21 @@ func TestOverridenTraceCall(t *testing.T) {
...
+		// 	call: ethapi.CallArgs{
+		// 		From:  &randomAccounts[0].addr,
+		// 		To:    &randomAccounts[1].addr,
+		// 		Value: (*hexutil.Big)(big.NewInt(1000)),
+		// 	},
+		// 	config: &TraceCallConfig{
+		// 		Tracer: &tracer,
+		// 	},
+		// 	expectErr: core.ErrInsufficientFundsForTransfer,
+		// 	expect:    nil,
```
meowsbits, line 378: Would it make sense to modify the test to `expectErr: nil; expect: <blah>`? (but leaving a comment about the difference, as you already have done)

ziogaschr: Thought of that too, but this test, wanna test the override functionality. Testing what you described, is being covered already in the tracers tests.

cmd/utils/flags.go
```diff
@@ -1738,37 +1738,36 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
 
-			isValid := false
-			for _, proto := range eth.DefaultProtocolVersions {
-				if proto == uint(version) {
-					isValid = true
-					seenVersions[uint(version)] = nil
-					break
-				}
+		isValid := false
+		for _, proto := range eth.DefaultProtocolVersions {
```
meowsbits: This validates the passed protocol version using `DefaultProtocolVersions` as the list of valid options (rather than as -- one might assume -- a list of _defaults_). Is this desired? Is there a list of eligibile-but-not-necessarily-default versions?

ziogaschr: Nice catch. Indeed the `eth.DefaultProtocolVersions` is misleading here. 
The intention was, that this value holds all the supported networks, the same way the used to have in the protocols.go (go-ethereum, till recently), and is being used as the Default value for the flag. 
So we have to either rename it or use a second var for it. 
I am on this one...

ziogaschr: Resolved by 15e0f16

PR #379, opened 2021-05-13
cmd/utils: set default for --eth.protocols flag

meowsbits:
The default is assigned to the config later if `len(config.protocols) == 0`, so this should not change any effect; but intention is to provide a more user-facing and consistent experience.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1771,11 +1771,6 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		}
 	}
 
-	// set default protocol versions
-	if len(cfg.ProtocolVersions) == 0 {
-		cfg.ProtocolVersions = eth.DefaultProtocolVersions
-	}
-
 	// Set DNS discovery defaults for hard coded networks with DNS defaults.
 	switch {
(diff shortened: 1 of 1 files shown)


ziogaschr:
Thanks. LGTM, and it is more readable, will test it and approve it.

meowsbits (commented):
See if https://github.com/etclabscore/core-geth/pull/379/files#R1775 should still exist; I think it _does no harm :registered:_, but may not be necessary.

ziogaschr:
> See if https://github.com/etclabscore/core-geth/pull/379/files#R1775 should still exist; I think it _does no harm ®️_, but may not be necessary.

This link doesn't point somewhere specifically.

ziogaschr (approved):
LGTM

meowsbits:
> > See if https://github.com/etclabscore/core-geth/pull/379/files#R1775 should still exist; I think it _does no harm registered_, but may not be necessary.
> 
> This link doesn't point somewhere specifically.

Ah, sorry. 


_flags.go line 1775_
```
// set default protocol versions
	if len(cfg.ProtocolVersions) == 0 {
		cfg.ProtocolVersions = eth.DefaultProtocolVersions
	}
```

ziogaschr:
indeed, this is not needed. I just tested it as well. Nice catch

PR #368, opened 2021-04-29
Set activation blocks for Magneto on classic|mordor|kotti chains

ziogaschr:
The ECIP https://github.com/ethereumclassic/ECIPs/pull/411 is still in draft mode. So I opened this PR with `A3-inprogress` label assigned.

It's complimentary to #365.

PR #385, opened 2021-05-25
Set activation blocks for Magneto on classic|mordor|kotti chains

ziogaschr:
The the activation numbers for Magneto based on the Dev's call outcome at https://github.com/ethereumclassic/ECIPs/issues/424

It's complimentary to #365.
Replaces #368.


meowsbits (approved):
Numbers LGTM.

meowsbits (changes requested):
Ah, wait. I think the `forkid`s need updating, too.

meowsbits (approved):
Added `forkid` tests with fbbf6e0.

PR #394, opened 2021-06-07
cmd/utils: add default snap discovery to config during setup

meowsbits:
This is behavior established at ethereum/go-ethereum for
now, though the snap-specific protocol DNS registrations
are disused; assumes coexistance of eth and snap protos.

Date: 2021-06-07 14:11:47-05:00


ziogaschr (approved):
LGTM. Let's keep our eyes open, for when they use the correct bootnodes at ethereum/go-ethereum or when we feel that a good percentage of nodes at ETC are running with snap on.

PR #398, opened 2021-07-13
Allow using --gcmode=archive and --txlookuplimit=0 flags

AusIV:
When running with --gcmode=archive, it automatically sets --txlookuplimit=0, but if you explicitly set --txlookuplimit=0 it raises an exclusivity exception.

This change will only raise the exclusivity exception if --gcmode=archive is set and --txlookuplimit is explicitly set to something other than zero, allowing both flags to be set consistently with other Geth derivatives.

This is especially useful to us at Rivet so we don't have to have special rules for ETC to exclude the --txlookuplimit flag (or, as has historically been the case, a fork with this check commented out).

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1591,8 +1591,8 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	CheckExclusive(ctx, DeveloperFlag, DeveloperPoWFlag, MainnetFlag, RopstenFlag, RinkebyFlag, GoerliFlag, YoloV3Flag, ClassicFlag, KottiFlag, MordorFlag, MintMeFlag)
 	CheckExclusive(ctx, LightServeFlag, SyncModeFlag, "light")
 	CheckExclusive(ctx, DeveloperFlag, DeveloperPoWFlag, ExternalSignerFlag) // Can't use both ephemeral unlocked and external signer
-	CheckExclusive(ctx, GCModeFlag, "archive", TxLookupLimitFlag)
 	if ctx.GlobalString(GCModeFlag.Name) == "archive" && ctx.GlobalUint64(TxLookupLimitFlag.Name) != 0 {
+		CheckExclusive(ctx, GCModeFlag, "archive", TxLookupLimitFlag)
 		ctx.GlobalSet(TxLookupLimitFlag.Name, "0")
 		log.Warn("Disable transaction unindexing for archive node")
 	}


cmd/utils/flags.go
```diff
@@ -1591,8 +1591,8 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	CheckExclusive(ctx, DeveloperFlag, DeveloperPoWFlag, MainnetFlag, RopstenFlag, RinkebyFlag, GoerliFlag, YoloV3Flag, ClassicFlag, KottiFlag, MordorFlag, MintMeFlag)
 	CheckExclusive(ctx, LightServeFlag, SyncModeFlag, "light")
 	CheckExclusive(ctx, DeveloperFlag, DeveloperPoWFlag, ExternalSignerFlag) // Can't use both ephemeral unlocked and external signer
-	CheckExclusive(ctx, GCModeFlag, "archive", TxLookupLimitFlag)
 	if ctx.GlobalString(GCModeFlag.Name) == "archive" && ctx.GlobalUint64(TxLookupLimitFlag.Name) != 0 {
+		CheckExclusive(ctx, GCModeFlag, "archive", TxLookupLimitFlag)
 		ctx.GlobalSet(TxLookupLimitFlag.Name, "0")
```
ziogaschr, line 1596: This will never be triggered. I suggest keeping either the `CheckExclusive` or set of `TxLookupLimit = 0`. @meowsbits what do you think?

AusIV: This will still be triggered if you explicitly set TxLookupLimit to a non-zero value.

AusIV: My thinking is that if somebody isn't setting TxLookupLimit but is setting gcmode=archive, the behavior of setting TxLookupLimit to zero (away from the default) is reasonable.

If someone is explicitly setting TxLookupLimit to zero, throwing an error because gcmode=archive was going to set it to zero anyway doesn't make sense.

But if somebody is trying to set TxLookupLimit to a value that is neither the default nor the value that gcmode=archive will set it to, then they probably want to know that's not a support configuration.

ziogaschr: I am sorry for the confusion @AusIV. I wanted to select the next line 1596. I agree that your PR is needed. What I am questioning/suggesting is that we either use `CheckExclusive` which will throw an error or we remove it and leave the next line to set it to `0` and continue.

AusIV: The current default value is non-zero. If you leave the default but use the archive flag, this will set it to zero.

I believe both lines are useful.

meowsbits: I like what AusIV has here; reusing `CheckExclusive` to error "normally" if the `TxLookupLimitFlag` value _isn't zero_ with `archive` mode (since archive mode necessitates/implies keeping all tx indexes).

ziogaschr: > The current default value is non-zero. If you leave the default but use the archive flag, this will set it to zero.
> 
> I believe both lines are useful.

Sorry guys, I forgot to reply. Indeed nice catch @AusIV. I checked and it works fine. I was looking at the code only without checking the default value.

ziogaschr (approved):

meowsbits (approved):
Thank you!

PR #400, opened 2021-07-22
Introduce MintMe hardfork to avoid contract address collisions

lukaszmatczak:
This pull request introduces support for MintMe mainnet network upgrade at height 252,500. It fixes problem with contract address collisions.

Network upgrade happened today.

### DIFF
--- a/params/types/coregeth/chain_config.go
+++ b/params/types/coregeth/chain_config.go
@@ -211,7 +211,7 @@ type CoreGethChainConfig struct {
 
 	RequireBlockHashes map[uint64]common.Hash `json:"requireBlockHashes"`
 
-	Lyra2NonceTransitionBlock *big.Int `json:"-"`
+	Lyra2NonceTransitionBlock *big.Int `json:"lyra2NonceTransitionBlock,omitempty"`
 }
 
 // String implements the fmt.Stringer interface.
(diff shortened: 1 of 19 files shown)


params/types/coregeth/chain_config.go
```diff
@@ -210,6 +210,8 @@ type CoreGethChainConfig struct {
 	BlockRewardSchedule         ctypes.Uint64BigMapEncodesHex `json:"blockReward,omitempty"`          // JSON tag matches Parity's
 
 	RequireBlockHashes map[uint64]common.Hash `json:"requireBlockHashes"`
+
+	Lyra2NonceTransitionBlock *big.Int `json:"-"`
```
meowsbits:
```suggestion
	Lyra2NonceTransitionBlock *big.Int `json:"lyra2NonceTransitionBlock,omitempty"`
```

I think you want this to be included in JSON. The configuration is stored with genesis in the database as JSON and fills the runtime configuration on startup if no flag is set. An unwritten value will cause its omission in this case.

This also applies to the other configuration structs.

params/types/goethereum/goethereum.go
```diff
@@ -77,6 +77,8 @@ type ChainConfig struct {
 
 	// Cache types for use with testing, but will not show up in config API.
 	ecbp1100Transition *big.Int
+
+	Lyra2NonceTransitionBlock *big.Int `json:"-"`
```
meowsbits:
```suggestion
	Lyra2NonceTransitionBlock *big.Int `json:"lyra2NonceTransitionBlock,omitempty"`
```

params/types/coregeth/chain_config.go
lukaszmatczak: You're right, thank you for noticing it.

meowsbits (approved):

Issue #404, opened 2021-08-22
Failed to read account password contents(faucet.exe)


zrock23:
faucet.exe....

←[32mINFO ←[0m[08-22|05:19:56.794] Using chain/net config                   ←[32mnetwork id←[0m=4 ←[32mbootnodes←[0m=enode://a24ac7c5484ef4ed0c5eb2d36620ba4e4aa13b8c84684e1b4aab0cebea2ae45cb4d375b77eab56516d34bfbd3c1a833fc51296ff084b770b94fb9028c4d25ccf@52.169.42.101:30303 ←[32mchain config←[0m="{ChainID: 4 Homestead: 1 DAO: <nil> DAOSupport: false EIP150: 2 EIP155: 3 EIP158: 3 Byzantium: 1035301 Constantinople: 3660663 ConstantinopleFix: 4321234 Istanbul: 5435345 MuirGlacier: <nil> Berlin: 8290928 Engine: ethash}"
←[35mCRIT ←[0m[08-22|05:19:56.808] Failed to read account password contents ←[35mfile←[0m= ←[35merr←[0m="open : The system cannot find the file specified."

what does this mean?
I do not understand anything significant anymore?

meowsbits:
The `faucet` program sends transactions (transferring funds). The faucet needs to use a key to sign these transactions (a keyfile for an account having some balance that you wish to distribute). For this, you need to use these two flags:

```
--account.json=/path/to/keyfile.json \
--account.pass=/path/to/password.txt
```

zrock23:
faucet.exe...
←[32mINFO ←[0m[08-24|18:59:07.420] Using chain/net config                   ←[32mnetwork id←[0m=4 ←[32mbootnodes←[0m=enode://a24ac7c5484ef4ed0c5eb2d36620ba4e4aa13b8c84684e1b4aab0cebea2ae45cb4d375b77eab56516d34bfbd3c1a833fc51296ff084b770b94fb9028c4d25ccf@52.169.42.101:30303 ←[32mchain config←[0m="{ChainID: 4 Homestead: 1 DAO: <nil> DAOSupport: false EIP150: 2 EIP155: 3 EIP158: 3 Byzantium: 1035301 Constantinople: 3660663 ConstantinopleFix: 4321234 Istanbul: 5435345 MuirGlacier: <nil> Berlin: 8290928 Engine: ethash}"
[...]

PR #405, opened 2021-08-23
cmd/faucet: add help notices and flag use audit for required account flags

meowsbits:
Rel #404 

Provides more useful help and in-case-of-error exit messages for these two required flags.


ziogaschr (approved):
LGTM

PR #413, opened 2021-09-21
Merge/foundation release/1.10.8

meowsbits:


ziogaschr (approved):
LGTM. Reviewed many times and tested if it syncs with most chains.

Our only concern was on some failing tests and EOA, which we conclude that it was a test issue which was fixed and not an actual issue in the client itself.

PR #415, opened 2021-09-21
.github/workflows,.travis.yml: migrate lint and a couple tests to Github Actions

meowsbits:
Travis has a new _Credit_ system that ~~coregeth~~ etclabscore/ is using up (1000 minutes / month) causing Travis to not run anymore until the new month rolls over.

This PR migrates the `make lint` check and a couple simple tests to Github Actions to remove some of the minutes from Travis.

It doesn't touch any Travis jobs that handle deploys (on tags). This is possible in the future, but I just want to get the easy stuff moved over first.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -1,4 +1,7 @@
 name: Lint and Tests
+env:
+  # Set to empty.
+  ANDROID_HOME:
 
 # Controls when the action will run.
 on:
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -3,6 +3,8 @@ name: Lint and Tests
 # Controls when the action will run.
 on:
   push:
+    branches:
+      - master
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -1,0 +1,59 @@
+name: Lint and Tests
+
+# Controls when the action will run.
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  # Run linter check. Only test code linters on latest version of Go.
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,59 @@
...
+        id: go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - run: make test
```
ziogaschr: Don't we need a `make all` here?

meowsbits: Thanks. Fixed with 4183ae5fae

meowsbits:
I would like to get the `pull_request` and `push` jobs run exclusively (eg. only run `pull_request` jobs when both hooks are triggered).

ziogaschr:
This will be nice and will save us some minutes as well from CI.

ziogaschr (approved):

ziogaschr:
I approved it, but let's first see if the tests pass here :) Was very fast to approve

meowsbits:
@ziogaschr Note the go1.17 bump

meowsbits:
I am working on this over here for now, to avoid spamming here.
https://github.com/meowsbits/core-geth/pull/21

meowsbits:
Despite the docs, `ANDROID_HOME` is set in Github Actions for the ubuntu-latest image.

https://docs.github.com/en/actions/learn-github-actions/environment-variables
https://github.com/meowsbits/ci-tests/runs/3689090391?check_suite_focus=true

I don't think it's set in Travis. 

When `ANDROID_HOME` isn't set and the `$HOME/Android/Sdk` path doesn't exist, the test is skipped.

I'm going to set the var to empty to force the test to skip. The directory doesn't exist.

meowsbits:
be advised that I force pushed

ziogaschr:
Great find man. I were looking on the wrong direction to solve this one. Thanks for saving us time.
I am fine skipping and already skipped test for android.

PR #427, opened 2021-10-29
Move release artifacts build from Travis to GitHub actions

ziogaschr:


ziogaschr:
@meowsbits can we remove these [travis tests](https://github.com/etclabscore/core-geth/blob/master/.travis.yml#L10-L31) as a similar test is being run [here](https://github.com/etclabscore/core-geth/blob/master/.github/workflows/test-linux.yml#L50-L66) (excluding coverage)?

ziogaschr:
@meowsbits feel free to create a tag and a draft release so as you can see it run

.github/workflows/release-packages.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  build-linux:
+    name: Build artifacts
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+        - os: macos-latest
+          BUILD_OS_NAME: macos
```
meowsbits: This was `osx` with Travis (eg. `core-geth-alltools-osx-v1.12.1.zip`). Is there a reason to use `macos` instead?

.github/workflows/release-packages.yml
```diff
@@ -0,0 +1,71 @@
...
+          image_additional_mb: 2000
+          copy_artifact_path: ./build/bin/
+          copy_artifact_dest: ./build/bin/
+          commands: |
+              wget https://golang.org/dl/go1.16.9.linux-armv6l.tar.gz
+              sudo tar -C /usr/local -xzf go1.16.9.linux-armv6l.tar.gz
+              rm go1.16.9.linux-armv6l.tar.gz
+              export PATH=$PATH:/usr/local/go/bin
+
+              GOARM=5 go run build/ci.go install -dlgo -arch arm
```
meowsbits, line 61: Should this be `6` to agree with the Go version installed just above?

.github/workflows/release-packages.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Prepare archive's for release
```
meowsbits:
```suggestion
      - name: Prepare archives for release
```

.github/workflows/release-packages.yml
meowsbits: (I generally think `macos` is a "better" name (since maybe one day in a distant galaxy Apple will finally make the major version bump... , and then OSX would be OSXI ;)
But just thinking about users' programmatic download concerns/dependencies.

ziogaschr: Nope, my bad. Nice catch man, multi cursor magic, while copying the name from the line above. Resolved by c4422acb

.github/workflows/release-packages.yml
ziogaschr: Tricky one for consistency.

`GOARM=5` is copied from https://github.com/etclabscore/core-geth/pull/427/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L78

The installed go above is used to run `go run build/ci.go`, while the `-dlgo` will install the correct needed version. It's a mess indeed. Will check if I can install a better go in the system for `GOARM=5`.

build/deploy.sh
```diff
@@ -1,14 +1,28 @@
 #!/usr/bin/env bash
 
-GETH_ARCHIVE_NAME="core-geth-${TRAVIS_OS_NAME}-$(git describe --always)"
-zip -j "$GETH_ARCHIVE_NAME.zip" build/bin/geth
 
-shasum -a 256 $GETH_ARCHIVE_NAME.zip
-shasum -a 256 $GETH_ARCHIVE_NAME.zip > $GETH_ARCHIVE_NAME.zip.sha256
+GETH_ARCHIVE_NAME="core-geth-${BUILD_OS_NAME}"
+ALLTOOLS_ARCHIVE_NAME="core-geth-alltools-${BUILD_OS_NAME}"
```
meowsbits: I think we still need to get the version identifier in here, eg. `v1.12.2`. 

`core-geth-alltools-arm.zip` should be `core-geth-alltools-arm-v1.12.2.zip`

.github/workflows/release-packages.yml
meowsbits: Roger. I know very little about ARM architecture versions. Is it worth publishing 5,6,whatever else?

build/deploy.sh
ziogaschr: Sorry, will add them back. I am sure I have seen somewhere the released zips being without the version.

.github/workflows/release-packages.yml
ziogaschr: I am not familiar with ARM arch too, we can publish them if we like, but maybe it makes sense to do once asked by the community.

build/deploy.sh
ziogaschr: Resolved by 690f1ed.
You can check a build [here](https://github.com/ziogaschr/core-geth/releases/tag/v1.12.12-ghbuilds)

meowsbits (approved):

PR #437, opened 2022-01-03
Tests/generate mystique

meowsbits:
Generates and includes `GeneralStateTests` for the `ETC_Mystique` test configuration.
This configuration is spec'd here: https://github.com/ethereumclassic/ECIPs/pull/452/files#diff-d5496c21eb89f07be78b9bbf3399a610defa66676e25cc5f803eef279a873b35.

I ran the generation program twice to provide some redundancy given potentially nondeterministic conditions (at least Go's un-guaranteed map ordering). The first of these is used for this pull request; the latter should be discarded after review.

https://github.com/etclabscore/core-geth/tree/tests/generate-mystique
https://github.com/etclabscore/core-geth/tree/tests/generate-mystique2


ziogaschr (approved):
LGTM

PR #440, opened 2022-01-08
Set Mystique activation numbers

ziogaschr:
based on ethereumclassic/ECIPs#458


meowsbits (approved):
LGTM. Would like to see https://github.com/etclabscore/core-geth/pull/437 merged first.

PR #448, opened 2022-01-13
Fix/arm releases

meowsbits:
Note that I want to merge it into the `release/*` branch. We already tagged and pushed v1.12.5, but it the CI build step was buggy. My plan is to move the tag ahead above this merge commit, and trigger a new build.


ziogaschr (approved):
LGTM

PR #447, opened 2022-01-12
Release/v1.12.5

meowsbits:

PR #449, opened 2022-01-13
Release/v1.12.6

meowsbits:
Follows/replaces #447. The CI ARM builds on the previous tag v1.12.5 were broken. That has been fixed thanks to @ziogaschr , and artifacts are now provided for ARMv5, v6, v7, and ARM64. 

Please note that in accommodation of these new artifacts, a (slightly) new naming pattern will be used: `-arm5-`, `-arm6-`, `-arm7-`, and `-arm64`. The `-arm-` artifact is now equivalent to `-arm5-`, and will not be included in release artifacts after this tagged release.


ziogaschr (approved):
LGTM

PR #452, opened 2022-01-31
Set the default derivation path for HD wallets on classic network

### LINKED ISSUE #451: HD wallet to have default DerivationPath for ETC apps `m/44'/61'/0'/0`
As a user, I would like to be able to load my HD wallet ETC addresses instead of the ETH ones.

When using Ledger it loads `m/44'/60'/0'/0` which has the addresses from the ETH app, while I would like the ETC app addresses to be loaded from `m/44'/61'/0'/0`.

#### System information

Geth version: 1.12.6
OS & Version: OSX

ziogaschr:
This PR suggests setting the default derivation path for HD wallets to point on ETC classic wallets instead of ETH ones.
This is a first direction on handling the derivation path, based on the fact that core-geth is mostly used for ETC networks.

Also I am starting the discussion here on wether we should make this configurable based on each chain we support at core-geth. Though it seems that nobody else used it (no issues so far), and we might put that in low prio.

Fixes #451 

List of canonical derivation paths: https://github.com/satoshilabs/slips/blob/master/slip-0044.md


meowsbits:
Hm, this is a tough one. 

A change from 60 to 61 as the default seems to me like a breaking API change.
Implementing configuration seems useful, but will have maintenance costs long term.

ziogaschr:
Indeed, it will have a lot of maintenance cost and I don't see this getting merged at foundation.

On the other hand, it's hard to use an HD wallet on ETC. What else I am thinking now, is that on every wallet invocation we will just print a warning that an ETH derivation path is used, with an example on how to use the correct one (if possible)

accounts/hd_test.go
```diff
@@ -30,43 +30,43 @@ func TestHDPathParsing(t *testing.T) {
...
-		{"m/44'/60'/0'/128", DerivationPath{0x80000000 + 44, 0x80000000 + 60, 0x80000000 + 0, 128}},
-		{"m/44'/60'/0'/0'", DerivationPath{0x80000000 + 44, 0x80000000 + 60, 0x80000000 + 0, 0x80000000 + 0}},
-		{"m/44'/60'/0'/128'", DerivationPath{0x80000000 + 44, 0x80000000 + 60, 0x80000000 + 0, 0x80000000 + 128}},
-		{"m/2147483692/2147483708/2147483648/0", DerivationPath{0x80000000 + 44, 0x80000000 + 60, 0x80000000 + 0, 0}},
-		{"m/2147483692/2147483708/2147483648/2147483648", DerivationPath{0x80000000 + 44, 0x80000000 + 60, 0x80000000 + 0, 0x80000000 + 0}},
+		{"m/44'/61'/0'/0", DerivationPath{0x80000000 + 44, 0x80000000 + 61, 0x80000000 + 0, 0}},
+		{"m/44'/61'/0'/128", DerivationPath{0x80000000 + 44, 0x80000000 + 61, 0x80000000 + 0, 128}},
+		{"m/44'/61'/0'/0'", DerivationPath{0x80000000 + 44, 0x80000000 + 61, 0x80000000 + 0, 0x80000000 + 0}},
+		{"m/44'/61'/0'/128'", DerivationPath{0x80000000 + 44, 0x80000000 + 61, 0x80000000 + 0, 0x80000000 + 128}},
+		{"m/2147483692/2147483709/2147483648/0", DerivationPath{0x80000000 + 44, 0x80000000 + 61, 0x80000000 + 0, 0}},
```
meowsbits, line 37: Just a note to self: 
- `fromhex(0x8000003d) => 2147483709`

https://github.com/satoshilabs/slips/blob/master/slip-0044.md

ziogaschr:
Closing this for now, later we might invest time on it, so as it is configurable with a flag or any other non breaking change.

PR #453, opened 2022-02-02
cmd/utils: fix --dev mode Fatalf for genesis conf

meowsbits:
Using --dev flag caused geth to
Fatalf. The program was trying to
install the genesis, where each
of two genesis resolution logics
fataled under dev mode. Apparently
dev mode does not use normal people's
genesis functions.

This seems to make it work.

Date: 2022-02-02 07:44:57-08:00


ziogaschr (approved):
LGTM and tested

PR #466, opened 2022-04-07
Ci/fix bench

meowsbits:
- Fixes the installation of `benchstat` dependency for benchmark output comparison.
- Fixes and un-skips two benchmark tests. See https://github.com/ethereum/go-ethereum/pull/24657.


ziogaschr:
Nice finding https://github.com/ethereum/go-ethereum/pull/24657 @meowsbits. 
PR LGTM.

ziogaschr (approved):

PR #467, opened 2022-04-14
Add native callTracer and refactor based on new go-ethereum paradigms

ziogaschr:
This PR achieves the following:
[x] Refactor of parity like tracer added at core-geth, so as it follows the go-ethereum paradigms.
[x] Add new **native** `callParityTracer` that is much faster as it runs in Go.
[x] Add more tests for tracers.
[x] Update the compatibility layer with OE based on their latest changes, and running a full comparison test based on this [OE branch](https://github.com/ziogaschr/openethereum/commits/etc) which used OE v3.3.4.

--

`stateDiff` tracer will be simplified and be converted in a native tracer in a next PR.


meowsbits:
Is this still a WIP? Any in particular you'd like feedback on?

ziogaschr:
No it's not, I just want to squash some commits.
I will also remove the new callTracerParityJs tracer as it is exactly the same as the native one and it only adds technical debt and adds an extra worry on how to remove it later on.

I suggest we merge this branch on Monday and continue with stateDiff on a new branch.

eth/tracers/internal/tracetest/testdata/call_tracer_parity/deep_calls.json
```diff
@@ -120,7 +120,8 @@
       },
       "blockNumber": 25001,
       "result": {
-        "gasUsed": "0x12bb3"
+        "gasUsed": "0x12bb3",
+        "output": "0x"
```
meowsbits, line 124: `0x`'s are in! Do docs need updating?

eth/tracers/internal/tracetest/testdata/call_tracer_parity/suicide.json
```diff
@@ -0,0 +1,117 @@
+{
```
meowsbits, line 1: How did you make this test data? Should we consider including any test creation tools in the repo?

meowsbits (approved):

eth/tracers/internal/tracetest/testdata/call_tracer_parity/deep_calls.json
ziogaschr: No we don't. They were in already under curtain conditions, while now it's the default, except some conditions.

eth/tracers/internal/tracetest/testdata/call_tracer_parity/suicide.json
ziogaschr: By using [this example code](https://github.com/etclabscore/core-geth/blob/d7543b9c21631466226e0f931edca13f92f9acca/eth/tracers/internal/tracetest/calltrace_test.go#L49-L96), but with the tracer name I need the test for.

Do you think we need to add anything more?

meowsbits: Ah, looks great. Sorry I didn't see that earlier.

meowsbits:
LGTM :rocket:

eth/tracers/internal/tracetest/testdata/call_tracer_parity/suicide.json
ziogaschr: I even missed it for long time mate :)

PR #471, opened 2022-05-10
Add BaseFee on genesis block for dev.pow chain

ziogaschr:
This fixes mining (panic), when `--dev.pow` flag is being used, by setting the BaseFee on the genesis block.


meowsbits:
Checked locally, fixes it for me.
Note that it will NOT work if you re-run on top of a failing database version from `master`; you need to remove and replace the datadir (or at least the genesis block entry). 
The `--dev.pow` program runs by default with only an in-mem db, so this is an edge case.

ziogaschr:
Indeed, but because `datadir` has to be set to persist the DB and also that `--dev.pow` is a newer addition, I though of not trying to fix existing chains.
What do you think of that?

meowsbits (approved):

PR #473, opened 2022-06-07
Ci/fix jenkins go version

meowsbits:
The Jenkins pipeline which handles regression tests started failing. I believe this is because of out-of-date Go versions on the EC2 instances. The failure is here: https://ci.etccore.in/blue/organizations/jenkins/core-geth-regression/detail/core-geth-regression/233/pipeline.

Current pipeline version:

```
go version go1.14.4 linux/amd64
```

I've encountered a similar error locally before and I was able to resolve it by upgrading my local Go version, so that's what I'm trying here.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ pipeline {
                     steps {
                         sh "curl -L -O https://go.dev/dl/go1.18.3.linux-amd64.tar.gz"
                         sh "sudo rm -rf /usr/bin/go && sudo tar -C /usr/local -xzf go1.18.3.linux-amd64.tar.gz"
-                        sh "sudo mv /usr/local/go/bin/* /usr/bin/"
+                        sh "sudo cp /usr/local/go/bin/* /usr/bin/"
                         sh "go version"
                         sh "make geth && ./build/bin/geth version"
                         sh "rm -rf ${GETH_DATADIR}-kotti"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -22,6 +22,11 @@ pipeline {
                 stage('Kotti') {
                     agent { label "aws-slave-m5-xlarge" }
                     steps {
+                        sh "curl -L -O https://go.dev/dl/go1.18.3.linux-amd64.tar.gz"
+                        sh "sudo rm -rf /usr/bin/go && sudo rm -rf /usr/local/go && sudo tar -C /usr/local -xzf go1.18.3.linux-amd64.tar.gz"
```
meowsbits, line 26: `/usr/bin/go` is the default Go installation. `/usr/local/go` is the dir where we extract the downloaded Go artifacts to. Even though its normally redundant to remove the latter, it is useful when the instance gets reused and straggling PATH or artifacts otherwise may exist. Jenkins maintains the EC2 instances for a short period following a run to save on restart-startup cost/time, so rapid pushes can be run on the same "machine." 

The following stanzas specify the expected Go executable artifacts precisely to provide better information for the reader.

meowsbits:
Also note that this branch name matches the pattern `origin/ci/*` which, along with `origin/master`, are the configured remote triggers for Jenkins. That's how I got the Jenkins tests to test here in the first place ;).

meowsbits:
Latest, passing Jenkins run: https://ci.etccore.in/blue/organizations/jenkins/core-geth-regression/detail/core-geth-regression/242/pipeline/

ziogaschr:
LGTM

PR #487, opened 2022-07-06
Feat/classic hd derivation path

meowsbits:
This PR builds from and replaces #452 (thanks @ziogaschr) which provided the pattern for configurable HD derivation path support. This branch has been rebased on master since then.

When `--usb` is used in conjunction with `--classic` or any testnet value (eg. `--sepolia`), the `accounts` package defaults will be configured accordingly for the HD derivation path logic and defaults.

Comments are left using Ethereum Classic values as examples.


ziogaschr:
Great work @meowsbits. I want to give it a test with Ledger before approving it.

ziogaschr:
@meowsbits just tested it with Ledger and it works great.

I am wondering * if it makes sense to use the ETHClassic Slip (0x3d) for its testnets (Mordor & Kotti). That way, UX will not be confusing. I wanted to try using the Ledger's ETC app (accounts) on Mordor and it wasn't working. I am totally fine merging this PR as it is, as it is inline with the spec.

> * SLIP-0044: I checked that Slip suggests CoinType=1 for all testnets.

Nice work.

meowsbits:
What about a flag that would override (or "define with -force"?) the value? 

-  `--usb.coin-type-id` or, 
- `--usb.slip0044-id` or, ...
- `--usb.pathid`, `--usb.path-component`, ...

ziogaschr:
> What about a flag that would override (or "define with -force"?) the value?
> 
> * `--usb.coin-type-id` or,
> * `--usb.slip0044-id` or, ...
> * `--usb.pathid`, `--usb.path-component`, ...

Nice solution. I vote towards `--usb.coin-type-id` or `--usb.pathid`, but I am leaving the decision to you.

meowsbits:
Last two commits add `--usb.pathid`.

ziogaschr:
Works great. Let's merge it @meowsbits.

PR #498, opened 2022-08-25
Fix tests on MacOS (darwin)

ziogaschr:


cmd/geth/consolecmd_cg_test.go
```diff
@@ -165,7 +167,10 @@ func TestGethStartupLogs(t *testing.T) {
 		},
 	}
 	for i, c := range cases {
-		t.Run(fmt.Sprintf("TestGethStartupLogs/%d: %v", i, c.flags), func(t *testing.T) {
+		// fix for darwin, where long test names fail with "The ipc endpoint is longer than %d characters."
```
ziogaschr: @meowsbits another way is to change the test name only `if rpc.max_path_size < len([]byte(strings.Join(c.flags, " "))))`, for this we have to export the `max_path_size` in `rpc`.

meowsbits: https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

If the error you're seeing is because of the IPC path length limit, then we can try to shorten that up by shortening the `datadir` path, since geth should use the default `<datadir>/<alt chain?>/geth/geth.ipc`.

But if the error is just because of the test names, I recommend reducing them to a simple `Sprintf("%d", i)` -- which seems like a reasonable thing to do anyway (shoving the flags in there was a bad idea). 

I'm going to push a commit, see whatcha think.

ziogaschr: The issue was because of the IPC path, which was long as part of it was based on test name.
I am going to test your commits

meowsbits (approved):

PR #510, opened 2022-12-14
Merge/foundation release/1.10.26

ziogaschr:
This merges go-ethereum tag v1.10.26 over our latest https://github.com/etclabscore/core-geth/tree/merge/foundation-release/1.10.23-tests-generate-fixes-statefix-difficulty-test-v2-ripemd-tx-debug-no-ripemd-hack-evmc-dimitrys-etc-tests

PR #512, opened 2023-01-10
Merge/foundation release/1.10.26 <- Remove EVMCv10 upgrade, minor fixes

meowsbits:


params/types/genesisT/gen_genesis.go
```diff
@@ -72,6 +76,24 @@ func (g *Genesis) UnmarshalJSON(input []byte) error {
 		BaseFee    *math.HexOrDecimal256                       `json:"baseFeePerGas"`
 	}
 	var dec Genesis
+	// We have to look at the raw input, decide what kind of configurator schema it's using,
```
ziogaschr, line 79: Are we going to keep the custom modified version of this generated file?

meowsbits: Yes. 

(And for posterity, some context on this question...)

See https://github.com/etclabscore/core-geth/pull/510/files#diff-82d9b07e427791ba46325cb31f353728eb444a3e48ef0c662ffc2d5a18d1de9c

We're adding a CI check checking if generated files (`go generate`) are up-to-date.
And _this_ generated file is modified (which normally is a no-no), but I don't see any clean ways around this. The modification handles disambiguation between potential configuration types (`ctypes.ChainConfig`) when unmarshaling. And unmarshaling the config happens every time on start up, for example, when the config is read from the db and maybe-merged with a new default configuration.

ziogaschr (approved):

PR #531, opened 2023-02-26
Merge/foundation release/1.11

meowsbits:
Merges ethereum/go-ethereum branch `release/1.11`  to core-geth `master`.

Significant changes include
- Ropsten is dead and gone; Sepolia replaces it as the canonical testnet.
- [Shangahai feature support](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md) is added;
  - Introducing time-based fork definitions
- `eth/68` is the new best-default protocol
- 2 `debug_` API methods gone, 4 new ones introduced.
- Pebble DB.
- `personal` API is hashly deprecated, requiring opt-back-in with `--rpc.enabledeprecatedpersonal`.
- Removed iOS and Android builds

Things I'd like to consider for this or next release
- [x] Remove all Parity, Aleth, and Multigeth configuration support. I don't think many people use this. Being abandoned projects, their chain spec schemas have long since not supported configurations for features post-Byzantium (so they could not be well-supported).

PR #535, opened 2023-04-11
Merge/foundation release/1.11 rm other client configs

meowsbits:
Resolving [a TODO at this comment](https://github.com/etclabscore/core-geth/pull/531#issue-1600094337). 
Merge base: https://github.com/etclabscore/core-geth/pull/531

This PR proposed removing support for configuration data types of MultiGeth, OpenEthereum/Parity, Aleth, and Pythereum. Only the first two were well supported at core-geth anyways, but all have become disused in some way and supporting configuration interchange with them just doesn't make that much sense anymore. For example, since about Istanbul new EIPs have not been supported, so the data type was basically a stand-in anyways.


core/types/gen_receipt_json.go
```diff
@@ -25,7 +25,7 @@ func (r Receipt) MarshalJSON() ([]byte, error) {
 		TxHash            common.Hash    `json:"transactionHash" gencodec:"required"`
 		ContractAddress   common.Address `json:"contractAddress"`
 		GasUsed           hexutil.Uint64 `json:"gasUsed" gencodec:"required"`
-		EffectiveGasPrice *hexutil.Big   `json:"effectiveGasPrice,omitempty"`
+		EffectiveGasPrice *big.Int       `json:"effectiveGasPrice"`
```
meowsbits: https://github.com/ethereum/go-ethereum/commit/87186148e0c4e46f7825e01cdcc57e833d1d3be6
> This change adds a struct field EffectiveGasPrice in types.Receipt. The field is present
in RPC responses, but not in the Go struct, and thus can't easily be accessed via ethclient

This change is related to an error I'm seeing locally at `TestT8n`, where the `effectiveGasPrice: null` pair is present but unexpected. 
This branch has recently done a `go generate ./...` command.

Simply adding `,omitempty` to the `gen_receipt` `MarshalJSON` method will fix it, but seems again like there might be a better way than editing the generated file... but in other Receipt contexts it seems we're identical to upstream. 

- https://github.com/ethereum/go-ethereum/commit/87186148e0c4e46f7825e01cdcc57e833d1d3be6#diff-2e925d78462a4abc9f45e63fdaa56401becf8dce096ac72297e1adc85511e02cR28

ziogaschr (approved):
LGTM

meowsbits:
Heads up on https://github.com/etclabscore/core-geth/pull/535/commits/07eff873d2d09e9db927797cb1e5da3c8eba5b8b; the test was passing for unknown reasons, but this should hopefully prevent it from failing for unknown reasons, too.

Resolving https://github.com/etclabscore/core-geth/pull/535#pullrequestreview-1381375856.

PR #541, opened 2023-05-15
Bump GitHub Action Ubuntu runner version

diega:
I also introduced the usage of [actions/cache](https://github.com/actions/cache) to speedup the building process. After adding this step the action execution went from 16 to 9 minutes.

### DIFF
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -20,14 +20,10 @@ jobs:
         - os: macos-latest
           BUILD_OS_NAME: osx
 
-        # Build on 18.04 (Bionic) because the OS GLIBC is bumped above 2.26 in later versions (eg. 20.04).
-        # This GLIBC version included and used by default in at least Amazon Linux 2 and Ubuntu Bionic.
-        # The C compiler used by default in the 18.04 runner produces an executable which is runnable
-        # on Ubuntu Disco and Focal (tested by meowsbits); it is forward compatible.
-        - os: ubuntu-18.04
+        - os: ubuntu-20.04
           BUILD_OS_NAME: linux
 
-        - os: ubuntu-18.04
+        - os: ubuntu-20.04
           BUILD_OS_NAME: arm
 
         - os: windows-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release-packages.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       matrix:
         include:
         - os: macos-latest
-          BUILD_OS_NAME: osx
+          BUILD_OS_NAME: macos
```
meowsbits: I think this is a good change, but only to note that this value is used in the artifact naming scheme so the generated release download links will change too.

`https://github.com/etclabscore/core-geth/releases/download/v1.12.11/core-geth-osx-v1.12.11.zip` becomes `https://github.com/etclabscore/core-geth/releases/download/v1.12.11/core-geth-macos-v1.12.11.zip`

meowsbits:
Awesssoooommee! Got tests somewhere demonstrating that the builds build?  (The workflow only runs on tags).

.github/workflows/release-packages.yml
diega: oh, good catch. I can revert it if you want. I just changed it b/c it gave me old vibes.

diega:
> Awesssoooommee! Got tests somewhere demonstrating that the builds build? (The workflow only runs on tags).

yes, I ran it on my fork. [This](https://github.com/diega/core-geth/actions/runs/4985581651) is the first run and [this](https://github.com/diega/core-geth/actions/runs/4985881714) one is the second, where you can see the usage of the cache.

.github/workflows/release-packages.yml
ziogaschr: I agree that it's better to not change the naming pattern on this one.

ziogaschr:
Great PR @diega. This will win us some executional minutes on GH as well.

.github/workflows/release-packages.yml
diega: cool, I just changed it back. Here's the [diff](https://github.com/etclabscore/core-geth/compare/da9d92ea78fd83983a522e75768deccf5aa58a00..5b6d627954c9e2ad8d0c22fce6caaa351133265c) after the rebase

ziogaschr (approved):
LGTM

meowsbits (approved):

PR #544, opened 2023-05-17
[zurich] Ghost-128 patch: downloader sync periodic TD check

meowsbits:
This patch applies the countermeasure
to security vulnerability Ghost-128 described
by Taverna and Paterson of ETH Zurich in their
2023 paper Snapping Snap Sync: Practical Attacks
on Synchronising Go-Ethereum Nodes.

This patch, in both its design and implementation,
was originally authored and provided by the authors
of that paper, and details around both can be
found in the paper's Section 7.

The patch has been edited only in adaptation
to minor application conflicts.

### DIFF
--- a/eth/downloader/downloader_test.go
+++ b/eth/downloader/downloader_test.go
@@ -158,9 +158,15 @@ type downloadTesterPeer struct {
 
 // Head constructs a function to retrieve a peer's current head hash
 // and total difficulty.
-func (dlp *downloadTesterPeer) Head() (common.Hash, *big.Int) {
+func (dlp *downloadTesterPeer) Head() (common.Hash, *big.Int, *big.Int) {
 	head := dlp.chain.CurrentBlock()
-	return head.Hash(), dlp.chain.GetTd(head.Hash(), head.Number.Uint64())
+	return head.Hash(), dlp.chain.GetTd(head.Hash(), head.Number.Uint64()), new(big.Int).Set(dlp.chain.CurrentBlock().Difficulty)
+}
+
+// SetHead constructs a function to retrieve a peer's current head hash
+// and total difficulty.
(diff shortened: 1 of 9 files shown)


diega (approved):
🚢

ziogaschr (approved):
Let's ship it

eth/downloader/downloader_test.go
```diff
@@ -986,8 +997,10 @@ func testHighTDStarvationAttack(t *testing.T, protocol uint, mode SyncMode) {
 	defer tester.terminate()
 
 	chain := testChainBase.shorten(1)
-	tester.newPeer("attack", protocol, chain.blocks[1:])
-	if err := tester.sync("attack", big.NewInt(1000000), mode); err != errStallingPeer {
+	dlp := tester.newPeer("attack", protocol, chain.blocks[1:])
+	fakeTD := big.NewInt(1000000)
+	dlp.fakeTD = fakeTD
```
meowsbits, line 1002: PTAL. This is from https://github.com/etclabscore/core-geth/pull/544/commits/e29e74f1dcb3f540cc9db5538782f29cc6e00037 which I'd like you to take an extra minute with.

Without this commit, the `*Starvation*` attacks fail. What I'm not 100% on is why... (and why this fixes it). 

Update: Here's why. The sync protocol calls `fetchHead` in `syncWithPeer`, where `fetchHead` assigns the `d.td` field. The tests were failing because the falsely-advertised TD was provided at the handshake, but the peer would honestly provide its TD (accessed from the chain data) during this early phase of sync.  
https://github.com/etclabscore/core-geth/pull/544/files#diff-80cc409bf7f1e4a3d19174ed878e4dcb3d40b19b6e0580bd3816a174c57e0fb2R748

PR #549, opened 2023-06-01
Rename RANDOM to PREVRANDAO in instruction set activation

ziogaschr:
During the merge from go-ethereum we wrongly handled the instruction to be activated for **RANDOM** which has been renamed to **PREVRANDAO**.

This can be found at https://github.com/ethereum/go-ethereum/commit/d30e39b2f833fb75f1e529cd405061fb6b548b8d#diff-3722385a96482d2250e1b71e50749b4a104968ca098dbb91331db536a37b3867L83


diega (approved):
good catch

PR #551, opened 2023-07-03
Merge/foundation release/1.12

ziogaschr:
Merges ethereum/go-ethereum branch release/1.12 to core-geth master.

In this merge we had to revert the changes made upstream with regards proof-of-work removal.
Also, core-geth now defaults to use **Pebble** as a backend if no existing database is found (https://github.com/ethereum/go-ethereum/pull/27136). If a previous LevelDB database exists Geth will keep using that, and if you must have LevelDB for some compatibility reasons, you can force it in Geth with the --db.engine=leveldb flag.


diega (approved):
Awesome! let's 🚢 it

PR #552, opened 2023-07-03
Drop Kotti support

ziogaschr:
Kotti is going to be deprecated soon. This PR drop support from core-geth client.


diega (approved):
Excellent. After the release, I'll do the same in Besu

PR #561, opened 2023-08-29
Merge/foundation release/1.12.2+spiral

meowsbits:
Implements the pending Spiral ([ECIP-1109](https://ecips.ethereumclassic.org/ECIPs/ecip-1109))  network upgrade for ETC.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -91,6 +91,8 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, triedb *trie.Database, gen
 
 	applyOverrides := func(config ctypes.ChainConfigurator) {
 		if config != nil {
+			// Block-based overrides are not provided because Shanghai is
+			// ETH-network specific and that protocol is defined exclusively in time-based forks.
 			if overrides != nil && overrides.OverrideShanghai != nil {
 				config.SetEIP3651TransitionTime(overrides.OverrideShanghai)
 				config.SetEIP3855TransitionTime(overrides.OverrideShanghai)
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -96,6 +96,8 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, triedb *trie.Database, gen
 
 	applyOverrides := func(config ctypes.ChainConfigurator) {
 		if config != nil {
+			// Block-based overrides are not provided because Shanghai is
+			// ETH-network specific and that protocol is defined exclusively in time-based forks.
 			if overrides != nil && overrides.OverrideShanghai != nil {
 				config.SetEIP3651TransitionTime(overrides.OverrideShanghai)
 				config.SetEIP3855TransitionTime(overrides.OverrideShanghai)
@@ -436,7 +438,7 @@ func GenesisToBlock(g *genesisT.Genesis, db ethdb.Database) *types.Block {
 	}
 	var withdrawals []*types.Withdrawal
 	if conf := g.Config; conf != nil {
-		isShangai := conf.IsEnabledByTime(g.Config.GetEIP4895TransitionTime, &g.Timestamp)
+		isShangai := conf.IsEnabledByTime(g.Config.GetEIP4895TransitionTime, &g.Timestamp) || g.Config.IsEnabled(g.Config.GetEIP4895Transition, new(big.Int).SetUint64(g.Number))
 		if isShangai {
 			head.WithdrawalsHash = &types.EmptyWithdrawalsHash
 			withdrawals = make([]*types.Withdrawal, 0)
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -438,8 +438,9 @@ func GenesisToBlock(g *genesisT.Genesis, db ethdb.Database) *types.Block {
 	}
 	var withdrawals []*types.Withdrawal
 	if conf := g.Config; conf != nil {
-		isShangai := conf.IsEnabledByTime(g.Config.GetEIP4895TransitionTime, &g.Timestamp) || g.Config.IsEnabled(g.Config.GetEIP4895Transition, new(big.Int).SetUint64(g.Number))
-		if isShangai {
+		// EIP4895 defines the withdwrawals tx type, implemented on ETH in the Shanghai fork.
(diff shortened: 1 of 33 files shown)


core/genesis.go
```diff
@@ -436,7 +438,7 @@ func GenesisToBlock(g *genesisT.Genesis, db ethdb.Database) *types.Block {
 	}
 	var withdrawals []*types.Withdrawal
 	if conf := g.Config; conf != nil {
-		isShangai := conf.IsEnabledByTime(g.Config.GetEIP4895TransitionTime, &g.Timestamp)
+		isShangai := conf.IsEnabledByTime(g.Config.GetEIP4895TransitionTime, &g.Timestamp) || g.Config.IsEnabled(g.Config.GetEIP4895Transition, new(big.Int).SetUint64(g.Number))
```
ziogaschr: Btw, I found out recently, that on the last release merge for some reason I used vars named as `isShangai` instead of `isEip4895`. I am thinking to revisit them and change them for consistency. What do you think guys?

core/vm/jump_table.go
```diff
@@ -193,7 +193,7 @@ func instructionSetForConfig(config ctypes.ChainConfigurator, isPostMerge bool,
 	if config.IsEnabled(config.GetEIP3198Transition, bn) {
 		enable3198(instructionSet) // BASEFEE opcode https://eips.ethereum.org/EIPS/eip-3198
 	}
-	if isPostMerge {
+	if isPostMerge || config.IsEnabled(config.GetEIP4399Transition, bn) { // EIP4399: Supplant DIFFICULTY opcode with PREVRANDAO (ETH @ PoS)
```
ziogaschr, line 196: Nice catch

eth/catalyst/api.go
```diff
@@ -169,7 +170,8 @@ func (api *ConsensusAPI) ForkchoiceUpdatedV1(update engine.ForkchoiceStateV1, pa
 		if payloadAttributes.Withdrawals != nil {
 			return engine.STATUS_INVALID, engine.InvalidParams.With(errors.New("withdrawals not supported in V1"))
 		}
-		if api.eth.BlockChain().Config().IsEnabledByTime(api.eth.BlockChain().Config().GetEIP4895TransitionTime, &payloadAttributes.Timestamp) {
+		if api.eth.BlockChain().Config().IsEnabledByTime(api.eth.BlockChain().Config().GetEIP4895TransitionTime, &payloadAttributes.Timestamp) || // PTAL(meowsbits) Is it really safe/sensible to call the backing Blockchain here?
+			api.eth.BlockChain().Config().IsEnabled(api.eth.BlockChain().Config().GetEIP4895Transition, api.eth.BlockChain().CurrentHeader().Number) {
```
ziogaschr: Note to me, to revisit it

params/types/coregeth/chain_config_configurator.go
```diff
@@ -516,51 +516,147 @@ func (c *CoreGethChainConfig) SetEIP4399Transition(n *uint64) error {
...
 	c.EIP6049FTime = n
+	if n != nil && *n == 0 {
+		c.EIP6049FBlock = big.NewInt(0)
+	}
+	return nil
+}
+
+// Shanghai by block
+// EIP3651: Warm COINBASE
+func (c *CoreGethChainConfig) GetEIP3651Transition() *uint64 {
```
ziogaschr, line 569: I like how the setters and getters of EIPs that can be activated with both time and block number check to each other, although I am a bit worried to what happens if both are being set by mistake and if the user will notice it.

Do you think that it is better if we take precedence of one over the other? For example, first check time then block number no matter which getter is being used.

Currently the siblings is only checked if enabled on genesis block, though some developers might mistakenly pass 0 values to some EIPs when they init chains using custom genesis.

meowsbits: > Do you think that it is better if we take precedence of one over the other? For example, first check time then block number no matter which getter is being used.

The problem is that each getter can only represent the time OR block value; a single getter can't reconcile between the two kinds of activation values (except for the special `0` case, where `time:0 -> block:0`; even `block:0 -> time:0` is not logically sound, since genesis blocks can, and do, have non-`0` timestamps).

And I think the problem persists, were precedence to be established, with developers mistakenly passing `nil` values (true zero-values), the same with `0` values. 

But I'm not sure I've perfectly understood your proposal, nor that I've even provided a watertight reason -- maybe you can show a pseudocode or snippet of what you're considering?

core/genesis.go
meowsbits: I like `isEIP4895` better, and even upstream is now following that pattern more and more. Comments about EIP::Forkname are always useful though )

ziogaschr: I will handle those in a different PR. Resolving this conversation now.

params/types/coregeth/chain_config_configurator.go
ziogaschr: @meowsbits I removed this logic as it's not needed. There was a copy/paste typo which caused all this. Here is the fix [26265c7](https://github.com/etclabscore/core-geth/pull/561/commits/26265c73c011f2e678f6aff4fc99db114cc5d4bf).

eth/catalyst/api.go
ziogaschr: Fixed by 87843da

ziogaschr (approved):
LGTM. Nice work man

ziogaschr:
Oh actually, before we merge this, shall we set the block numbers for Spiral correctly? 
Currently classic is set to `18_400_000`, while Mordor is not at all.

meowsbits:
Ah, important. I think we should unset all activation numbers for ETC until the ECIP gets more love and confidence. (ECIP 1109 currently specs impossible (past) values of `18000000` for mainnet, `9000000` for Mordor); https://github.com/ethereumclassic/ECIPs/blob/64423c73b7416c2f8e40d4c66bf40b715c458b8c/_specs/ecip-1109.md).

meowsbits:
Test was failing, but is another flaky one https://github.com/etclabscore/core-geth/actions/runs/6084737553/job/16507292035?pr=561#step:4:1218. Cannot reproduce locally. Merging.

PR #580, opened 2023-11-02
Forkid tests

meowsbits:
A small addition/fix to the `GatherForks` tests. Before, only block-based forks were checked. Now, also check time-based forks. Also add test cases for default networks Goerli and Sepolia.


ziogaschr (approved):
LGTM

Issue #567, opened 2023-10-08
My nodes are unable to synchronize


YuXiaoCoder:
#### System information

Geth version: `1.12.14-stable-f1c80f7e`
OS & Version: Linux

#### Expected behaviour
Node Normal Synchronization

#### Actual behaviour
Nodes cannot be synchronized, showing that they cannot connect to Peer

#### Steps to reproduce the behaviour
geth --classic --datadir=/data --syncmode=snap --gcmode=archive --http --http.addr=0.0.0.0 --http.port=8545 --http.vhosts=* --ws --ws.addr=0.0.0.0 --ws.port=8546

#### Backtrace

````
INFO [10-08|11:22:58.356] Started P2P networking                   self=enode://268e827ac0c8f7069a18a02fb5cf4985c377b35e1c383485707de09847788bb3d3bc41fe0a6dd4f5d351cf75ca33b6410fc14fbe1ca906c99101855426a3b7ad@127.0.0.1:30303
INFO [10-08|11:22:58.356] WebSocket enabled                        url=ws://127.0.0.1:8551
INFO [10-08|11:22:58.356] HTTP server started                      endpoint=127.0.0.1:8551 auth=true  prefix= cors=localhost vhosts=localhost
INFO [10-08|11:22:58.403] Generated state snapshot                 accounts=8893 slots=0 storage=409.64KiB dangling=0 elapsed=88.751ms
INFO [10-08|11:23:08.369] Looking for peers                        peercount=0 tried=14 static=0
INFO [10-08|11:23:09.980] Block synchronisation started
WARN [10-08|11:23:09.981] Synchronisation failed, retrying         err="peer is unknown or unhealthy"
INFO [10-08|11:23:18.453] Looking for peers                        peercount=0 tried=29 static=0
INFO [10-08|11:23:22.581] New local node record                    seq=1,696,735,378,353 id=69347bd3fbdd0b43 ip=8.218.35.139 udp=30303 tcp=30303
WARN [10-08|11:23:23.966] Synchronisation failed, retrying         err="peer is unknown or unhealthy"
INFO [10-08|11:23:28.701] Looking for peers                        peercount=0 tried=33 static=0
INFO [10-08|11:23:49.036] Looking for peers                        peercount=2 tried=38 static=0
INFO [10-08|11:24:09.361] Looking for peers                        peercount=1 tried=35 static=0
INFO [10-08|11:24:19.365] Looking for peers                        peercount=1 tried=36 static=0
[...]

PR #582, opened 2023-11-02
Investigate/peering

meowsbits:
Reports like #567 and others in other channels (mostly Discord) suggest some kind of vague issue with peering. This is a WIP to explore the problem and potential solutions.


ziogaschr:
@meowsbits what do you think of merging this PR and start using the snap DNS list?

meowsbits:
Thanks for the bump on this. I just rebased on master and pushed. There were 4 subsequent commits that I hadn't pushed, but now have. I'd like to freshen up on what's going on here and then I'll come back to ya, especially since it seems like we've addressed probably the brunt of the peering issues with your #611.

ziogaschr:
We can close this one I think

PR #591, opened 2023-11-22
Add Besu bootnodes

diega:


diega:
Not needed for the moment

PR #592, opened 2023-11-29
Disable MESS

meowsbits:
This PR drafts the disablement of  [MESS/ECBP1100](https://ecips.ethereumclassic.org/ECIPs/ecip-1100), anticipated to occur on ETC concurrently with the [Spiral](https://ecips.ethereumclassic.org/ECIPs/ecip-1109) network upgrade. 

Following convention, a configuration override flag `--ecbp1100.disable=[number]` is provided.

As-is, this draft uses `runtime` and `reflect` at the `Configurator` implementation level to handle the actual enable/disable logic (https://github.com/etclabscore/core-geth/commit/9e274d02851888a9dab51d5ba197193f09aa2b24). Alternatively, we could consider instead adding conditions at the places where the configuration is referenced to handle it, like the example below.

```diff
diff --git a/core/forkchoice.go b/core/forkchoice.go
index d07cdf6aa4..55d898c3a0 100644
--- a/core/forkchoice.go
+++ b/core/forkchoice.go
@@ -162,7 +162,8 @@ func (f *ForkChoice) ReorgNeeded(current *types.Header, extern *types.Header) (b
                        return reorg, nil
                }
        }
-       if !f.chain.Config().IsEnabled(f.chain.Config().GetECBP1100Transition, current.Number) {
+       if !f.chain.Config().IsEnabled(f.chain.Config().GetECBP1100Transition, current.Number) ||
+               f.chain.Config().IsEnabled(f.chain.Config().GetECBP1100DisableTransition, current.Number) {
                return reorg, nil
        }

```

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1068,7 +1068,7 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Usage: "Configure ECBP-1100 (MESS) block activation number",
 		Value: math.MaxUint64,
 	}
-	ECBP1100DisableFlag = &cli.Uint64Flag{
+	ECBP1100DeactivateFlag = &cli.Uint64Flag{
 		Name:  "ecbp1100.disable",
 		Usage: "Disable ECBP-1100 (MESS) block activation number",
 		Value: math.MaxUint64,
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1068,6 +1068,11 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Usage: "Configure ECBP-1100 (MESS) block activation number",
 		Value: math.MaxUint64,
 	}
+	ECBP1100DisableFlag = &cli.Uint64Flag{
+		Name:  "ecbp1100.disable",
+		Usage: "Disable ECBP-1100 (MESS) block activation number",
+		Value: math.MaxUint64,
+	}
 	ECBP1100NoDisableFlag = &cli.BoolFlag{
 		Name:  "ecbp1100.nodisable",
 		Usage: "Short-circuit ECBP-1100 (MESS) disable mechanisms; (yields a permanent-once-activated state, deactivating auto-shutoff mechanisms)",
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2086,6 +2086,8 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 			cfg.EthDiscoveryURLs = SplitAndTrim(urls)
 		}
 	}
+	CheckExclusive(ctx, ECBP1100DisableFlag, ECBP1100NoDisableFlag)
+
 	// Override any default configs for hard coded networks.
 
 	// Override genesis configuration if a --<chain> flag.
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1064,18 +1064,19 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Value: "",
 	}
 	ECBP1100Flag = &cli.Uint64Flag{
-		Name:  "ecbp1100",
-		Usage: "Configure ECBP-1100 (MESS) block activation number",
-		Value: math.MaxUint64,
+		Name:     "ecbp1100", // should have been override.ecbp1100, but maintained now for backwards compatibility
+		Usage:    "Manually specify the ECBP-1100 (MESS) block activation number, overriding the bundled setting",
+		Category: flags.EthCategory,
 	}
-	ECBP1100DeactivateFlag = &cli.Uint64Flag{
-		Name:  "ecbp1100.disable",
-		Usage: "Disable ECBP-1100 (MESS) block activation number",
-		Value: math.MaxUint64,
+	OverrideECBP1100DeactivateFlag = &cli.Uint64Flag{
+		Name:     "override.ecbp1100.deactivate",
+		Usage:    "Manually specify the ECBP-1100 (MESS) deactivation block number, overriding the bundled setting",
+		Category: flags.EthCategory,
 	}
 	ECBP1100NoDisableFlag = &cli.BoolFlag{
-		Name:  "ecbp1100.nodisable",
-		Usage: "Short-circuit ECBP-1100 (MESS) disable mechanisms; (yields a permanent-once-activated state, deactivating auto-shutoff mechanisms)",
+		Name:     "ecbp1100.nodisable",
+		Usage:    "Short-circuit ECBP-1100 (MESS) disable mechanisms; (yields a permanent-once-activated state, deactivating auto-shutoff mechanisms)",
+		Category: flags.DeprecatedCategory,
 	}
 
 	MetricsEnableInfluxDBV2Flag = &cli.BoolFlag{
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2086,7 +2086,6 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 			cfg.EthDiscoveryURLs = SplitAndTrim(urls)
 		}
 	}
-	CheckExclusive(ctx, ECBP1100DisableFlag, ECBP1100NoDisableFlag)
 
 	// Override any default configs for hard coded networks.
 
--- a/params/config_classic.go
+++ b/params/config_classic.go
@@ -81,9 +81,9 @@ var (
 		EIP2028FBlock: big.NewInt(10_500_839),
 		EIP2200FBlock: big.NewInt(10_500_839), // RePetersburg (=~ re-1283)
 
-		ECBP1100FBlock:        big.NewInt(11_380_000), // ETA 09 Oct 2020
-		ECBP1100DisableFBlock: big.NewInt(69_420_123),
-		ECIP1099FBlock:        big.NewInt(11_700_000), // Etchash (DAG size limit)
+		ECBP1100FBlock:           big.NewInt(11_380_000), // ETA 09 Oct 2020
+		ECBP1100DeactivateFBlock: big.NewInt(69_420_123),
+		ECIP1099FBlock:           big.NewInt(11_700_000), // Etchash (DAG size limit)
 
 		// Berlin eq, aka Magneto
 		EIP2565FBlock: big.NewInt(13_189_133),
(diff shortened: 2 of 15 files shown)


params/types/coregeth/chain_config_configurator.go
```diff
@@ -669,6 +681,13 @@ func (c *CoreGethChainConfig) IsEnabled(fn func() *uint64, n *big.Int) bool {
 	if f == nil || n == nil {
 		return false
 	}
+	fnName := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
```
ziogaschr, line 684: Nice hack in order to check if both are enabled, either way it will not stay for too long.

p.s.: I hope I don't read back this comment in 1-2 years and the code is still here :)

ziogaschr:
I wonder if we also have to handle cases where `--ecbp1100.nodisable` is set.

Shall we force disable and raise a warning? This is what we prefer, on the other way, it's against the user's preference.
I think if user decided to use `--ecbp1100.nodisable` we can only show a warning but we can't force a disable.

cmd/utils/flags.go
```diff
@@ -2081,6 +2086,8 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 			cfg.EthDiscoveryURLs = SplitAndTrim(urls)
 		}
 	}
+	CheckExclusive(ctx, ECBP1100DisableFlag, ECBP1100NoDisableFlag)
```
ziogaschr: @meowsbits this is probably better to be removed (I added it).

A node operator might indeed want the `nodisable` functionality, so if he still wants to enable ECBP-1100 he has to do `--ecbp1100.disable=69420123 --ecbp1100.nodisable`.

The thing is that with the above check, this is not allowed.

meowsbits:
Good point on the `nodisable` flag. That feature was on request, and was intended to turn off the MinPeers and StaleHead safety mechanisms. The concern was that the safety mechanisms could be used to knock a node off the canonical network, assuming that the canonical network followed MESS. If was MESS was on, they wanted it to _stay_ on. 

> MESS is only enabled if a peer has greater than or equal to the MinimumSyncPeers peers. In Core-Geth this value is by default 5.
MESS is disabled if, once synced, a node's head block is not changed within a time limit (ie becomes stale). In Core-Geth this value is by default 30 * 13 seconds
> https://github.com/ethereumclassic/ECIPs/blob/master/_specs/ecip-1100.md#implementation

So its my understanding that conceptually there's no conflict with these flags, although you're definitely right to point to it.

ziogaschr:
Oh, so this will not allow MESS to be disabled only once it has been enabled. 
Is this right? If yes, then there is no conflict indeed.

params/config_classic.go
```diff
@@ -81,8 +81,9 @@ var (
 		EIP2028FBlock: big.NewInt(10_500_839),
 		EIP2200FBlock: big.NewInt(10_500_839), // RePetersburg (=~ re-1283)
 
-		ECBP1100FBlock: big.NewInt(11_380_000), // ETA 09 Oct 2020
-		ECIP1099FBlock: big.NewInt(11_700_000), // Etchash (DAG size limit)
+		ECBP1100FBlock:        big.NewInt(11_380_000), // ETA 09 Oct 2020
+		ECBP1100DisableFBlock: big.NewInt(69_420_123),
```
diega: What does `69_420_123` mean?

diega (approved):
🚢

params/config_classic.go
ziogaschr: Just far in future. I will not be surprised if it has a meaning though :)

meowsbits: Yea its just a placeholder block number.

meowsbits:
The naming/conceptual conflict of `ecbp1100.nodisable` and `GetECBP1100DisableTransition` is annoying and confusing. 

IMO `ecbp1100.nodisable` could have been better named `ecbp1100.nosafety`, but now is unchangeable since it holds an API place in the flags and config TOML marshalling:

```go
	ECBP1100NoDisableFlag = &cli.BoolFlag{
		Name:  "ecbp1100.nodisable",
		Usage: "Short-circuit ECBP-1100 (MESS) disable/safety mechanisms; (yields a permanent-once-activated state, deactivating auto-shutoff mechanisms)",
	}
`````

```go
	// ECBP1100NoDisable overrides
	// When this value is *true, ECBP100 will not (ever) be disabled; when *false, it will never be enabled.
	ECBP1100NoDisable *bool `toml:",omitempty"`
```

This configuration value is used as:

```go
func (bc *BlockChain) EnableArtificialFinality(enable bool, logValues ...interface{}) {
	// Short circuit if AF state is enabled and nodisable=true.
	if bc.artificialFinalityNoDisable != nil && atomic.LoadInt32(bc.artificialFinalityNoDisable) == 1 &&
		bc.IsArtificialFinalityEnabled() && !enable {
		log.Warn("Preventing disable artificial finality", "enabled", true, "nodisable", true)
		return
	}
```

How about renaming our new `GetECBP1100DisableTransition` and her associated fields to `ECBP1100Deactivate` instead to lessen the confusion? "Activation" being common language for turning on fork features ("block activation number"), and we're talking about "turning off" the feature.

params/types/coregeth/chain_config.go
```diff
@@ -184,8 +184,9 @@ type CoreGethChainConfig struct {
 	ECIP1017EraRounds  *big.Int `json:"ecip1017EraRounds,omitempty"` // ECIP1017 era rounds
 	ECIP1080FBlock     *big.Int `json:"ecip1080FBlock,omitempty"`
 
-	ECIP1099FBlock *big.Int `json:"ecip1099FBlock,omitempty"` // ECIP1099 etchash HF block
-	ECBP1100FBlock *big.Int `json:"ecbp1100FBlock,omitempty"` // ECBP1100:MESS artificial finality
+	ECIP1099FBlock        *big.Int `json:"ecip1099FBlock,omitempty"`        // ECIP1099 etchash HF block
+	ECBP1100FBlock        *big.Int `json:"ecbp1100FBlock,omitempty"`        // ECBP1100:MESS artificial finality
+	ECBP1100DisableFBlock *big.Int `json:"ecbp1100DisableFBlock,omitempty"` // Disable ECBP1100:MESS artificial finality
```
ziogaschr:
```suggestion
	ECBP1100DeactivateFBlock *big.Int `json:"ecbp1100DeactivateFBlock,omitempty"` // Deactivate ECBP1100:MESS artificial finality
```

params/config_classic.go
```diff
@@ -81,8 +81,9 @@ var (
 		EIP2028FBlock: big.NewInt(10_500_839),
 		EIP2200FBlock: big.NewInt(10_500_839), // RePetersburg (=~ re-1283)
 
-		ECBP1100FBlock: big.NewInt(11_380_000), // ETA 09 Oct 2020
-		ECIP1099FBlock: big.NewInt(11_700_000), // Etchash (DAG size limit)
+		ECBP1100FBlock:           big.NewInt(11_380_000), // ETA 09 Oct 2020
+		ECBP1100DeactivateFBlock: big.NewInt(69_420_123),
```
ziogaschr: Let's not forget to set the block_number here or we wait for this PR to be merged and set it at #595

params/types/coregeth/chain_config.go
```diff
@@ -184,8 +184,9 @@ type CoreGethChainConfig struct {
 	ECIP1017EraRounds  *big.Int `json:"ecip1017EraRounds,omitempty"` // ECIP1017 era rounds
 	ECIP1080FBlock     *big.Int `json:"ecip1080FBlock,omitempty"`
 
-	ECIP1099FBlock *big.Int `json:"ecip1099FBlock,omitempty"` // ECIP1099 etchash HF block
-	ECBP1100FBlock *big.Int `json:"ecbp1100FBlock,omitempty"` // ECBP1100:MESS artificial finality
+	ECIP1099FBlock           *big.Int `json:"ecip1099FBlock,omitempty"`                 // ECIP1099 etchash HF block
+	ECBP1100FBlock           *big.Int `json:"ecbp1100FBlock,omitempty"`                 // ECBP1100:MESS artificial finality
+	ECBP1100DeactivateFBlock *big.Int `json:"ecbp1100DeactivateFBlockFBlock,omitempty"` // Disable ECBP1100:MESS artificial finality
```
ziogaschr:
```suggestion
	ECBP1100DeactivateFBlock *big.Int `json:"ecbp1100DeactivateFBlockFBlock,omitempty"` // Deactivate ECBP1100:MESS artificial finality
```

params/types/goethereum/goethereum_configurator.go
```diff
@@ -691,6 +703,13 @@ func (c *ChainConfig) IsEnabled(fn func() *uint64, n *big.Int) bool {
 	if f == nil || n == nil {
 		return false
 	}
+	fnName := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
+	if strings.Contains(fnName, "ECBP1100Transition") {
+		disabledTransition := c.GetECBP1100DeactivateTransition()
```
ziogaschr: @meowsbits  this is the last one. `s/disabledTransition/deactivateTransition/g`

meowsbits: Danke, done and force pushed again (squashed into https://github.com/etclabscore/core-geth/pull/592/commits/7a026c079bf33bceca4629cb8132628f99baad41)

params/types/coregeth/chain_config.go
meowsbits: I made it and squashed into https://github.com/etclabscore/core-geth/pull/592/commits/e4b7dd2f1b2ebcfe7c62f29e22bc4eebfee1881d

params/types/coregeth/chain_config.go
meowsbits: Done and squashed into https://github.com/etclabscore/core-geth/pull/592/commits/e4b7dd2f1b2ebcfe7c62f29e22bc4eebfee1881d

params/config_classic.go
meowsbits: I just set it here == Spiral https://github.com/etclabscore/core-geth/pull/592/commits/907295dcdbcc66bf4474eae22f3341c0941ea3b9

params/config_classic.go
meowsbits: Configured to == Spiral w/ https://github.com/etclabscore/core-geth/pull/592/commits/907295dcdbcc66bf4474eae22f3341c0941ea3b9

diega (approved):
🚢²

ziogaschr (approved):

PR #595, opened 2023-12-06
Set Spiral classic activation block

ziogaschr:
ECIP Spec: https://github.com/ethereumclassic/ECIPs/pull/517


diega (approved):
let's the party begin 🎊

meowsbits (approved):

PR #598, opened 2023-12-07
cmd/devp2p: avoid exceeding cloudflare record quota

### LINKED ISSUE #323: handle cloudflare record quota better
The cloudflare record quota is 3500. 

This failure https://github.com/etclabscore/discv4-dns-lists/issues/322 is caused by exceeding it (or attempting to). https://github.com/etclabscore/discv4-dns-lists/actions/runs/7127904738/job/19408807629#step:13:20

We attempted to fix it with https://github.com/etclabscore/discv4-dns-lists/pull/314 by truncating the raw list of peers to push to 3500, but that didn't work.

It didn't work because if we have 3500 records existing on CF, and want to push 3500 where some of those are new, then we'll exceed the quota because the `devp2p dns to-cloudflare` command will attempt to push new and updated records before deleting stale ones.

If we naively modify the deploy command to delete the stale records first, we could potentially create a negative space in the records before filling it, which could be risky if the record create/update command fails for any reason.

meowsbits:
Attempts to resolve https://github.com/etclabscore/discv4-dns-lists/issues/323

This patch modifies the deployment
pattern by deleting one stale record
before creating a new record, rather
than leaving all the stales to be deleted
at the end.

Date: 2023-12-07 08:24:07-07:00

### DIFF
--- a/cmd/devp2p/dns_cloudflare.go
+++ b/cmd/devp2p/dns_cloudflare.go
@@ -125,16 +125,41 @@ func (c *cloudflareClient) uploadRecords(name string, records map[string]string)
 		}
 		existing[strings.ToLower(entry.Name)] = entry
 	}
+	// if the record exists on cloudflare, but is not in our local records, it is stale
+	stales := make(map[string]cloudflare.DNSRecord)
+	for path, entry := range existing {
+		if _, ok := records[path]; !ok {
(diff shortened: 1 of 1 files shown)


ziogaschr (approved):
Great addition

PR #602, opened 2024-01-08
params: install ECBP1110 block 10_400_000 (MESS deactivation)

### LINKED ISSUE #601: deactivate MESS on Mordor
[ECBP-1110](https://github.com/meowsbits/ECIPs/blob/ecbp1110/_specs/ecip-1110.md) specifies MESS's deactivation for ETC, but not Mordor. But Mordor should probably deactivate it too -- thanks [@chris-mercer for the flag on this](https://github.com/etclabscore/core-geth/pull/597#issuecomment-1863647918). 

We'll just need to settle on a block number and add configuration here: https://github.com/etclabscore/core-geth/blob/master/params/config_mordor.go#L96.
Reference for deactivation on Classic: https://github.com/etclabscore/core-geth/blob/master/params/config_classic.go#L85

meowsbits:
Resolves https://github.com/etclabscore/core-geth/issues/601.

Block `10,400,000` is only ~4 days way, so this'll need to be quick if we want to actually get Mordor participants to deactivate simultaneously, although, being a testnet, it's not critical. There's around 4 miners on Mordor currently.

### DIFF
--- a/params/config_mordor.go
+++ b/params/config_mordor.go
@@ -88,12 +88,14 @@ var (
 		// EIP4895FBlock: nil, // Beacon chain push withdrawals as operations
 		EIP6049FBlock: big.NewInt(9_957_000), // Deprecate SELFDESTRUCT (noop)
 
-		DisposalBlock:      big.NewInt(0),
-		ECIP1017FBlock:     big.NewInt(0),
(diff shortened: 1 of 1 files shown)


diega (approved):
Excellent

ziogaschr (approved):

PR #611, opened 2024-02-01
Fix UPnP unwanted termination

### LINKED ISSUE #609: Why was UDP closed in the latest update?
Peers connecting has become horrible since the last update and i noticed the closed connection. it is effecting the entire network.
Why and what are we supposed to do about it?

ziogaschr:
This pull request fixes the issue where the portMappingLoop function was prematurely terminated, causing the defer function to be called and trigger "Deleting port mapping" 2 minutes after the initial launch. The fix involves changing the return statement to a continue statement in order to continue the loop.

Fixes #609.

Background: this has been introduced with https://github.com/ethereum/go-ethereum/pull/26359. If this PR solves the issue we will make a PR at [ethereum/go-ethereum](https://github.com/ethereum/go-ethereum/) as well.

### DIFF
--- a/p2p/server_nat.go
+++ b/p2p/server_nat.go
@@ -127,7 +127,7 @@ func (srv *Server) portMappingLoop() {
 			} else if !ip.Equal(lastExtIP) {
 				log.Debug("External IP changed", "ip", extip, "interface", srv.NAT)
 			} else {
-				return
+				continue
 			}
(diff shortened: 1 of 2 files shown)


diega (approved):
Astonishing catch!

meowsbits (approved):
:heart_eyes_cat:

PR #630, opened 2024-04-27
respect locals-settings

PutraLaksmana:
This change adds a check to ensure that transactions added to the legacy pool are not treated as 'locals' if the global locals-management has been disabled.

This change makes the pool enforce the --txpool.pricelimit setting.


diega:
Thank your for your contribution. This change is already merged into [upstream](https://github.com/ethereum/go-ethereum/pull/28435) and will be part of `core-geth` in the upcoming release.

PR #631, opened 2024-04-29
Update algorithm.go

yuriy0803:
Set the number of threads to 2  ?????????


meowsbits:
Why?

yuriy0803:
> Why?

because it is the next DAG size where the CPU runs at 100% Generating ethash verification cache

yuriy0803:
What other programs can cause problems until it generates the Ethash verification cache?

yuriy0803:
when CPU is running at 100%

ziogaschr:
While I see your point, as you see 100% CPU usage for a while, I wouldn't favour this change. The Ethash generation takes short time (1-5 minutes), which is considered ok to utilise the max number of CPU threads available to the system. The system handles its resources for other apps.

meowsbits:
It seems this proposal has been resolved as unnecessary by @ziogaschr's clarification.

PR #635, opened 2024-05-20
p2p: fix race in dialScheduler

meowsbits:


ziogaschr:
Close this as it's already in master.

PR #637, opened 2024-06-06
Fix lint issue SA1019 reflect.SliceHeader

ziogaschr:


diega (approved):
It's always good to have one less _reflect_ dependency 🤗

PR #652, opened 2025-01-09
chore: make function comments match function names

dashangcun:
make function comments match function names


diega:
I'm sorry, we have a new rule that we reject PRs that only contain typo fixes, because of many PR and airdrop farmers

PR #653, opened 2025-01-17
Drop support for Goerli network

ziogaschr:
Ethereum has already removed support of Goerli network. This PR removes support from the core-geth client as well.


diega (approved):
<3

PR #667, opened 2025-02-28
refactor: Directly swap variable values

shenpengfeng:
Directly exchange the values ​​of two variables to avoid introducing temporary variables.


diega:
Thanks for the patch. We work upstream-first here: core-geth is a fork of go-ethereum, so anything that isn't Ethereum Classic-specific should go to `ethereum/go-ethereum` first, and we pick it up from there.

There's nowhere to send this one, though. Upstream reworked `cmd/devp2p/internal/ethtest` and the `Addresses` helper is gone there. What's left on our side is test tooling, and the change is style-only. Going to pass.

PR #681, opened 2025-06-06
Fix space in //go:noescape directive

gcmutator:
Fix unnecessary space in the "//go:noescape" directive to resolve ineffectual compiler directive warning.


diega:
Thanks, this is a real one. Without the `//` attached, the compiler reads it as an ordinary comment and the directive never applies, so escape analysis has to assume `gfpNeg`'s pointer arguments escape. It sits on the BN256 pairing path, so it shows up under the EIP-197 precompiles.

Upstream hit the same thing and fixed it in ethereum/go-ethereum#29222. We work upstream-first, so we'll take it as a cherry-pick of their commit 4e1116f9c rather than merging this separately. Same change, and it keeps our history against upstream clean.

Closing here. The fix is going in.

PR #688, opened 2025-11-09
rpc: improve IPC socket error message for filesystem compatibility

### LINKED ISSUE #687: Show detail error information on IPC binding failed
It's not possible to use an exFAT partition as data folder, because the IPC socket on exFAT is not supported.

When I use a exFAT mobile driver, following line will throw an error:

    l, err := net.Listen("unix", endpoint)

Change to following might be more clear:

    if err != nil {
        return nil, errors.Join(err, fmt.Errorf("network: unix, address: %s, make sure the partition file system is supported(ext2/ext3/ext4,etc.)", endpoint))
    }

cloorc:
Clarify that ext2/ext3/ext4 filesystems support Unix domain sockets (including Thunderbolt Type-C drives), while exFAT does not. This helps users understand why socket creation may fail on certain filesystems.

This will fix #687

### DIFF
--- a/rpc/ipc_unix.go
+++ b/rpc/ipc_unix.go
@@ -50,7 +50,7 @@ func ipcListen(endpoint string) (net.Listener, error) {
 	os.Remove(endpoint)
 	l, err := net.Listen("unix", endpoint)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to listen on unix socket: %w; if the error persists, verify that the disk partition supports Unix domain sockets (ext2/ext3/ext4 work; exFAT does not)", err)
 	}
 	os.Chmod(endpoint, 0600)
 	return l, nil


diega:
Thanks for this, and sorry it sat for so long.

The diagnosis is a good one. A socket that fails because the partition has no support for Unix domain sockets is genuinely hard to work out from the bare error.

Same answer as #689 though: we work upstream-first, and this one applies to go-ethereum unchanged. Their `rpc/ipc_unix.go` still has the same bare `return nil, err`. Send it there, and if it lands we'll take it from upstream.

Two things worth tightening before you do. `net.Listen` returns a `*net.OpError` that already prints `listen unix /path: bind: operation not supported`, so the "failed to listen on unix socket" prefix repeats what the error already says. And naming three filesystems that work plus one that doesn't is more than the message needs. Something closer to this has better odds:

```go
return nil, fmt.Errorf("%w (the filesystem may not support Unix domain sockets)", err)
```

The exFAT detail is the useful part and reads better as a code comment than inside the error string.

Closing here for now.

PR #691, opened 2025-11-10
ci: upgrade blst to fix appveyer issue

### LINKED ISSUE #690: Incorrect macro result in appveyer continuous integration fail
For example: https://ci.appveyor.com/project/meowsbits/core-geth/builds/53045406

In file included from C:\Users\appveyor\go\pkg\mod\github.com\supranational\blst@v0.3.11\bindings\go\blst.go:16:
C:/Users/appveyor/go/pkg/mod/github.com/supranational/blst@v0.3.11/bindings/blst.h:27:15: error: 'bool' cannot be defined via 'typedef'
   27 | typedef _Bool bool; /* it's assumed that cgo calls modern enough compiler */
      |               ^~~~

In github.com/supranational/blst@v0.3.11/bindings/blst.h

```c
#ifdef __cplusplus
extern "C" {
#elif defined(__BLST_CGO__)
typedef _Bool bool; /* it's assumed that cgo calls modern enough compiler */
#elif defined(__STDC_VERSION__) && __STDC_VERSION__>=199901
# define bool _Bool
#else
# define bool int
#endif
```

The first version to change this macro is v0.3.14:

```c
#ifdef __cplusplus
extern "C" {
#elif !defined(__STDC_VERSION__) || __STDC_VERSION__<202311
# if defined(__BLST_CGO__)
typedef _Bool bool; /* it's assumed that cgo calls modern enough compiler */
# elif !defined(bool)
#  if defined(__STDC_VERSION__) && __STDC_VERSION__>=199901
#   define bool _Bool
#  else
#   define bool int
#  endif
#  define __blst_h_bool__
# endif
#endif
```

The later directives are more reasonable.

cloorc:
This PR will fix #690

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,7 +67,7 @@ require (
 	github.com/spf13/cobra v1.5.0
 	github.com/status-im/keycard-go v0.2.0
 	github.com/stretchr/testify v1.8.4
-	github.com/supranational/blst v0.3.11
+	github.com/supranational/blst v0.3.14
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/tidwall/gjson v1.6.0
 	github.com/tyler-smith/go-bip39 v1.1.0
--- a/go.sum
+++ b/go.sum
@@ -662,8 +662,8 @@ github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5
 github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.8.4 h1:CcVxjf3Q8PM0mHUKJCdn+eZZtm5yQwehR5yeSVQQcUk=
 github.com/stretchr/testify v1.8.4/go.mod h1:sz/lmYIOXD/1dqDmKjjqLyZ2RngseejIcXlSw2iwfAo=
-github.com/supranational/blst v0.3.11 h1:LyU6FolezeWAhvQk0k6O/d49jqgO52MSDDfYgbeoEm4=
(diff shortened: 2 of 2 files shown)


diega:
Thanks for this, your diagnosis is right. blst 0.3.11 defines `bool` unconditionally, and the MinGW toolchain on AppVeyor's image now defaults to C23, where it is a keyword. v0.3.14 guards that behind `__STDC_VERSION__ < 202311`, which is exactly the fix.

We applied it onto current master and verified it: `go mod tidy` moves nothing else, the full build passes, and `crypto/kzg4844` is green.

For the record this is not a critical path for us. blst only comes in through c-kzg-4844, and the point evaluation precompile that reaches it is registered only when EIP-4844 is enabled, which no chain config in this tree does. So the bump changes which toolchains can build the binary, not what the binary does.

One thing before merging, and it is about discipline rather than risk. We try to keep divergence from upstream as low as we can, in the code and in the git history alike, so that any change can be traced back mechanically. Upstream made the identical bump in [#31165](https://github.com/ethereum/go-ethereum/pull/31165), commit `24ed0b506602cf18e92c2ec5b75173a68178748b`. Would you mind replacing your commit with a cherry-pick of it?

```
git fetch https://github.com/ethereum/go-ethereum.git
git cherry-pick -x 24ed0b506602cf18e92c2ec5b75173a68178748b
```

`-x` adds the trailer and keeps Marius as author. Happy to do it ourselves if you prefer, either way the fix is yours to have found.

diega:
Thanks for this, and for the diagnosis in particular. You were right about the cause: blst 0.3.11 defines `bool` unconditionally, and the MinGW toolchain on AppVeyor's image now defaults to C23, where it is a keyword.

The direction we are taking with this repository is to follow upstream's changes rather than carry our own where upstream has already solved something, and to bring them in as upstream's own commits so the provenance stays visible. #702 does that for the Go toolchain as a whole: it walks blst forward through go-ethereum's own commits, reaching 0.3.16, which is what upstream ships in v1.17.5.

So we are closing this one, with the fix landing through that route. Your diagnosis is what pointed at the right version, and it would be good to have your eyes on #702 if you have time.

PR #693, opened 2026-02-23
اتریوم

jovi1355:
سلام ساخت آن یوم


diega:
This looks like it was opened by accident: it adds a new workflow file containing the single word `space`, which would not be a valid GitHub Actions configuration. Closing on that basis, with no problem at all if it was unintentional.

If you were meaning to propose a change to the project, please open a new pull request describing what it should do, and we will take a look.

PR #698, opened 2026-04-28
Implement ECIP-1049: Turned on Keccak

antsankov:


diega:
Thanks for putting this together, and apologies it has sat here.

Checking the current state of the specification before acting on it: ECIP-1049 was moved to Withdrawn in [ethereumclassic/ECIPs#486](https://github.com/ethereumclassic/ECIPs/pull/486), merged on 20 May 2022, and Besu removed its Keccak256 implementation and the associated test network the following year in [besu-eth/besu#5371](https://github.com/besu-eth/besu/pull/5371), citing that withdrawal as the reason.

With the specification withdrawn and no other client carrying an implementation, merging a Keccak256 consensus engine would leave core-geth maintaining a proof of work path with no network to serve, so the right thing is to close this.

You are a co-author of the ECIP, so this is very likely already known to you. Worth noting for the record that ECIP-1000 leaves the door open on your side: champions may move a proposal between Draft, Deferred and Withdrawn at their own discretion. If the proposal is ever taken back to Draft and finds a path forward, this branch is a solid starting point and we would be glad to look at it again.

PR #701, opened 2026-09-15
internal/debug, cmd/geth: drop the memsize debug endpoint

shrikus:
`core-geth` does not link on any Go toolchain from 1.23 onwards:

```
link: github.com/fjl/memsize: invalid reference to runtime.stopTheWorld
```

Go 1.23 started rejecting pull-`//go:linkname` references to runtime symbols that are not push-linknamed, and `github.com/fjl/memsize` reaches into the runtime to stop and start the world around its heap scan.

This matters now that Go 1.21 has been out of support since August 2024: the tree cannot currently be built on a maintained toolchain.

### Why not `-checklinkname=0`

The linker flag silences the check, but the check is reporting a real ABI break rather than a false positive. memsize v0.0.2 declares:

```go
//go:linkname stopTheWorld runtime.stopTheWorld
func stopTheWorld(reason stwReason)

//go:linkname startTheWorld runtime.startTheWorld
func startTheWorld()
```

while the runtime now has:

```go
func stopTheWorld(reason stwReason) worldStop
func startTheWorld(w worldStop)
```

`startTheWorld` gained a required parameter. Linking with the check disabled would leave it restarting the world from whatever happens to occupy the argument slot, so anyone hitting the `/memsize/` endpoint would get undefined behaviour instead of a link error.

### What this changes

`memsize` backed exactly one debug-only HTTP handler, `/memsize/`, registered on the pprof server, plus the `debug.Memsize.Add("node", stack)` call that fed it. Both are removed along with the dependency. go-ethereum dropped the same package upstream for the same reason.

No consensus, p2p, EVM or RPC code is touched — `/memsize/` was never part of the JSON-RPC surface, only the pprof HTTP server.

### Verification

Built and run with Go 1.27.1 on linux/amd64, and still builds on Go 1.21.

The workflows still pin Go 1.21, so CI will keep testing on the old toolchain; bumping those pins is a separate decision and is deliberately left out of this PR.


diega:
Thanks for this. You identified the same thing upstream did: the `//go:linkname` in `fjl/memsize` is what pins the toolchain, and removing the endpoint is the way out.

The direction we are taking with this repository is to follow upstream's changes rather than carry our own where upstream has already solved something, and to bring them in as upstream's own commits so the provenance stays visible. #702 does that for the Go toolchain as a whole, and it includes go-ethereum's own removal of memsize, [ethereum/go-ethereum#30253](https://github.com/ethereum/go-ethereum/pull/30253), under its original authorship.

You noted that the CI version pins were deliberately out of scope here. #702 covers those too, walking them up through upstream's commits to what v1.17.5 runs.

So we are closing this one, with the change landing through that route. It would be good to have your eyes on #702 if you have time, especially since you run this in production.

